PR #1, opened 2014-12-04
Remove SRFI-1 dependency

gus-massa:
This removes the use of the function `fold-right`, so it's not more necessary to require `srfi/1`. I think that this also remove the dependency to the `srfi-lite-lib` package, but I'm not sure how to test that.


mflatt:
Merged - thanks!
Running `raco setup` checks the dependency declaration and confirms (with reasonable certainty) that your change is right.

PR #2, opened 2015-01-09
Use hash instead of hasheq to manage gui models for typed racket compat

lexi-lambda:
This is a change to accommodate rackunit gui usage from typed racket. See racket/typed-racket#8.


samth:
@rmculpepper Is this ok to merge?

rmculpepper:
Yes, looks fine to me.

PR #3, opened 2015-07-14
Allow internal definitions in custom checks

jackfirth:
Currently `define-check` and friends do not allow internal definitions, so custom checks like the following fail:

```racket
(define-check (some-complicated-check arg ...)
  (define some-var (some-expr arg ...))
  (use-some-var-to-determine-failure ...))
```

This PR replaces uses of `(begin expr ...)` with `(let () body ...)` to allow internal definitions, and updates the docs to use the `body` metavariable instead of `expr`.


samth:
This looks nice. @rmculpepper?

stamourv:
Merged, thanks!

PR #6, opened 2015-08-19
Allow long check-info keys

jackfirth:
This change allows for longer check-info keys. Previously there was a hard maximum of 12 spaces allotted for keys, so if a key was longer than that, it got cut off:

```
some-key:   "some-value"
some-super-l"bklajdk"
```

With this change however, long keys push the values to the right, ensuring a minimum of one space between the colon-suffixed keys and values:

```
some-key:            "some-value"
some-super-long-key: ”bklajdk"
```

The minimum of 12 characters is preserved, so when given only short keys spacing stays the same as before:

```
a:          "some-value"
b:          "some-other-value"
```

### DIFF
--- a/rackunit-lib/rackunit/private/format.rkt
+++ b/rackunit-lib/rackunit/private/format.rkt
@@ -19,7 +19,7 @@
 ;; name-width : integer
 ;;
 ;; Number of characters we reserve for the check-info name column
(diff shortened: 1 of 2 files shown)


stamourv:
Merged, thanks!

PR #12, opened 2015-10-25
Adjust docs on checks and current-check-handler

takikawa:
I think the current docs are misleading on checks and wrong about the default value of `current-check-handler`. This PR should fix that.

Can someone else check if these adjustments are accurate?

Prompted by http://stackoverflow.com/questions/33324304/racketunit-check-is-not-throwing-exception


stamourv:
These changes look consistent with the implementation of checks, unlike the current prose.
LGTM

PR #15, opened 2016-04-16
moved rackunit/log to new pkg, fixed docs

florence:
Moving this to a new package lets packages like `redex` and `cover` depend on it without depending on all of rackunit.


bennn:
Does this mean redex will eventually stop saying "Both tests passed" ? (I hope not)

florence:
@bennn Nope! It just means that `redex` will report test counts and failures to `rackunit` (anything that uses those counts like `raco cover` and `raco test`).

rfindler:
Wow, I am glad someone noticed! :)

mflatt:
@florence A minor point, but is it ok if I change the author to "florence@racket-lang.org" to follow a weak convention of using racket-lang.org addresses for main-distribution packages?

samth:
I don't think `florence@racket-lang.org` exists -- I think we mostly stopped adding new racket-lang.org addresses when everything moved to github.

mflatt:
It looks to me like that address does exist.

florence:
I didn't know that email was a thing, but I can confirm it reaches me. Do I change the author to "florence@racket-lang.org", or just 'florence?

mflatt:
Just `'florence`

PR #16, opened 2016-06-24
test-case and test-begin should not catch ctrl-c's

florence:


rfindler:
+1

Also: not-beak-exn?

florence:
@rfindler `exn:not-break?` is in unstable it seems

rfindler:
Wow, I guess that was only in my imagination?

rfindler:
Maybe exn:fail?

florence:
I think a `(raise 5)` should cause your test to fail, but not kill the rest of your tests.

rfindler:
Ok by me.

florence:
@rmculpepper Can I merge this for the release?

rmculpepper:
Seems reasonable to me.

florence:
I pushed as  2407b02

PR #19, opened 2016-10-14
Update doc for check-exn: predicate can be trueish

ohspite:
The predicate in `check-exn` is a match if it returns a true value, not just `#t`.   

``` racket
(check-exn (λ (ex) #t) (thunk (error))) ; passes
(check-exn (λ (ex) 'trueish) (thunk (error))) ; passes
```

It's a minor thing, but the difference is worth noting since there are other checks where non-`#t` means failure (the predicate in `check-match`, for example).


LeifAndersen:
LGTM

PR #21, opened 2016-12-23
make the top-level checks look more like how they look when using text-ui

rfindler:
Or, in other words, try to make these two print similarly:

  #lang racket
  (require rackunit rackunit/text-ui)
  (check-equal? 1 2)
  (run-tests (test-suite "name" (check-equal? 1 2)))

### DIFF
--- a/rackunit-lib/rackunit/private/format.rkt
+++ b/rackunit-lib/rackunit/private/format.rkt
@@ -1,7 +1,9 @@
 #lang racket/base
 (require racket/match
          "base.rkt"
-         "check-info.rkt")
+         "check-info.rkt"
+         "text-ui-util.rkt"
+         "location.rkt")
 
 (provide display-check-info-name-value
          display-check-info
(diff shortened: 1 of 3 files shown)


rfindler:
FWIW, here is how they print without this pull request's changes:

```
--------------------
FAILURE
actual:     1
expected:   2
name:       check-equal?
location:   (#<path:/Users/robby/tmp.rkt> 3 0 50 18)
expression: (check-equal? 1 2)

Check failure
--------------------
--------------------
name > #f
Unnamed test 
FAILURE
name:       check-equal?
location:   /Users/robby/tmp.rkt:4:30
actual:     1
expected:   2
Check failure
--------------------
```

and with the pull request's changes the first one prints like this (and the second is unchanged):

```
--------------------
FAILURE
name:       check-equal?
location:   /Users/robby/tmp.rkt:3:0
actual:     1
expected:   2

Check failure
--------------------
```

jeapostrophe:
Great, thanks!

PR #22, opened 2017-05-10
Added screenshot of the RackUnit GUI, fixed typo.

jsmaniac:
Should I leave the program inside the screenshot, or should I trim it to only contain the RackUnit window?


stamourv:
I'd just show the Rackunit window, since the code shows up right after in typeset form. I'd probably show the code first, though.

jsmaniac:
@samth Updated, thanks for the input!

stamourv:
@samth and I are in fact different people. I'm the better-looking of the two. ;)
Merged, thanks!

jsmaniac:
Whoops, that was me typing faster than my own fingers. I knew shouldn't have installed auto-correct on my brain :) .

PR #23, opened 2017-05-22
Use define-runtime-path for the rackunit screenshot.

jsmaniac:
I just realized that images included with an immediate path in scribble can cause problems when the document is incuded via `@include-section` into a document in another package. This is unlikely to happen, but to be on the safe side this patch uses `define-runtime-path`.

Can someone who is more familiar with scribble confirm that this is the correct fix / best practice?


bennn:
Merged in 5bde07fc4908dc6e862cab2f3e0e3f6848b9eef9

PR #25, opened 2017-06-03
Omit check failure message when custom message not provided.

### LINKED ISSUE #11: Command line output includes noise
When run in DrRacket, a failing test includes clickable error buttons pointing to the source location of the failing test and showing the stacktrace of the error, in addition to a "Check failure" message for the error. When run at the command line, this output is given:

```
--------------------
FAILURE
name:       check
location:   some-location-data
expression: (check string=? "foo" "bar")
params:     (#<procedure:string=?> "foo" "bar")

Check failure
--------------------
```

The extra whitespace line and "Check failure" message seems unnecessary. Some custom checks customize the error message, but the built in checks don't. Would it be better to either have the default checks include human-readable messages with more information, to not print those two lines at the command line at all, or to remain as they are now?

jackfirth:
Closes #11 

After this change, the expression `(check-equal? 1 2)` prints an error message like this:

```
--------------------
FAILURE
name:       check-equal?
location:   unsaved-editor:3:0
actual:     1
expected:   2
expression: (check-equal? 1 2)
--------------------
```

A mild refactoring in `check.rkt` involving wrapper functions is included.

### DIFF
--- a/rackunit-lib/rackunit/private/check.rkt
+++ b/rackunit-lib/rackunit/private/check.rkt
@@ -72,7 +72,7 @@
          marks
          (check-info-stack marks)))))
     ((_)
-     (fail-check "Check failure"))))
+     (fail-check ""))))
 
 (define-syntax fail-internal
   (syntax-rules ()
--- a/rackunit-lib/rackunit/private/check.rkt
+++ b/rackunit-lib/rackunit/private/check.rkt
@@ -38,7 +38,7 @@
 ;; parameter current-check-handler : (-> any any)
 (define current-check-handler
   (make-parameter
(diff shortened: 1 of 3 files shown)


jeapostrophe:
I think you should merge it

AlexKnauth:
Did this pull request get rid of the "stop sign" icons in DrRacket that you could use to jump to the location of the failure?

Before this commit it looked like this:
check-failure-stop-sign

And after it looked like this:
check-failure-no-stop-sign

Did the "stop sign" go away when you removed the "Check failure" text?

jackfirth:
It did, see #42. I intend to fix that before the next release.

Issue #10, opened 2015-10-23
Nested check info?


jackfirth:
Suppose I have a situation where I want to register a logically related group of check infos. Currently there are two ways to do this:
1. Group them into a hash and register a single check info with the name describing what the group is

``` racket
--------------------
FAILURE
name:             some-check
location:         (unsaved-editor 12 0 126 5)
expression:       (foo)
params:           ()
some-extra-info:  #hash((a . 1) (b . 2))
--------------------
```
1. Register them all at the top level

``` racket
--------------------
FAILURE
name:       some-check
location:   (unsaved-editor 12 0 126 5)
expression: (foo)
params:     ()
a:          1
b:          2
--------------------
```

1) is undesirable to me because it's difficult to read, especially if there's more than two or three keys. 2) is undesirable to me because now I've lost the information that these are logically grouped keys, and I can't name them by group.

Ideally I would like some way to have a check failure say this:

``` racket
--------------------
FAILURE
name:             some-check
location:         (unsaved-editor 12 0 126 5)
expression:       (foo)
params:           ()
some-extra-info:
  a:              1
  b:              2
--------------------
```

Would this be useful to others? Is there a better way I should attempt to do this?

jeapostrophe:
Go for it! I suggest that you have the single some-extra-info field but it's value is a (make-rackunit-table X) where X is the hash, then during formatting it looks for this structure and prints it nice.

PR #27, opened 2017-06-03
Add string-info wrapper for info messages

### LINKED ISSUE #26: String check info without quotes
I'm working on a testing library and I wanted to add a check info value that is a string but displays without quotes. I propose a `(string-check-info str)` struct that RackUnit looks for when displaying check info, so `(with-check-info (['foo (string-check-info "some message")]) ...)` adds check info that prints as `foo:    some message` instead of `foo:    "some message"`. Similar to #10.

jackfirth:
Closes #26

With this change, it’s now possible for a check info value to print a message without quotes by wrapping the value string in a call to `string-info`. See the example added to the docs.

### DIFF
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -267,7 +267,7 @@ the @racket[with-check-info*] function, and the
 @racket[with-check-info] macro.
 
 @defstruct[check-info ([name symbol?] [value any])]{
- A check-info structure stores information associated with the context of
+ A check-info structure stores information associated with the context of the
  execution of a check. The @racket[value] is written in a check failure message
  using @racket[write] unless it is a @racket[string-info] value.}
 
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -279,7 +279,8 @@ the @racket[with-check-info*] function, and the
    #:eval rackunit-eval
    (with-check-info (['value "hello world"]
                      ['message (string-info "hello world")])
-     (check = 1 2)))}
+     (check = 1 2)))
+ @history[#:added "1.2"]}
 
 The are several predefined functions that create check
 information structures with predefined names.  This avoids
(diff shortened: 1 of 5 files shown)


jackfirth:
Pinging @rmculpepper for review, on recommendation of @stamourv

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -267,9 +267,20 @@ the @racket[with-check-info*] function, and the
 @racket[with-check-info] macro.
 
 @defstruct[check-info ([name symbol?] [value any])]{
-
-A check-info structure stores information associated
-with the context of execution of a check.}
+ A check-info structure stores information associated with the context of
```
bennn: should say "the context of the" ?

jackfirth:
That could work fine for this use case, but I don't think it would work as a solution to #10 and I was hoping to add a very similar feature like `(nested-info <check-info> ...)` that displays as an indented list of infos. That would be ugly to do with `gen:custom-write`, and I'd prefer a single solution that works for both of those use cases. I'm also considering a `(repeated-info <value> ...)` feature that makes `(check-info 'foo (repeated-info 1 2 3)` print like this:

```
---------------
FAILURE
foo:         1
foo:         2
foo:         3
---------------
```

I'm not yet sure if that's necessary for the extension library I'm making, but it _definitely_ isn't possible to do with `gen:custom-write`.

bennn:
Ok that makes more sense

(Does scribble have a text back-end we could re-use?)

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: Fixed

jackfirth:
Do you mean some sort of Scribble subsystem for handling how Rackunit prints check info? I don't know much about Scribble's internals but it sounds unlikely.

bennn:
I mean, a Scribble renderer that will convert `nested` structs into indented text and `itemize` structs into bullet-point text.

jackfirth:
There might be something like that in Scribble, but I wouldn't want to hold up this PR on figuring that out.

jeapostrophe:
I think this should be merged.

jackfirth:
Alas I don't have merge powers

PR #30, opened 2017-06-10
Fix check failure message tests

jackfirth:
#25 failed to properly update the tests and caused a [DrDr failure](http://drdr.racket-lang.org/40705/racket/share/pkgs/rackunit-test/tests/rackunit/standalone.rkt).


bennn:
Hooray looks fixed:
http://drdr.racket-lang.org/40730/

PR #34, opened 2017-06-11
Remove undocumented unused exports of text-ui

jackfirth:
A Github search failed to find any callers of these undocumented exports of rackunit/text-ui.

### DIFF
--- a/rackunit-lib/info.rkt
+++ b/rackunit-lib/info.rkt
@@ -11,4 +11,4 @@
 
 (define pkg-authors '(ryanc noel))
 
-(define version "1.3")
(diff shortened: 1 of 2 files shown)


jackfirth:
Bumped version and sent out a [mailing list post](https://groups.google.com/forum/#!topic/racket-users/XL0l90yI8LM) looking for users. How long do you think we should wait for responses?

bennn:
I think, 10 days total

jackfirth:
Rebased and ready. No responses from the mailing list post. Will merge once CI passes, then update post in case anything breaks.

PR #37, opened 2017-06-14
Simplify strip-redundant-params implementation

jackfirth:
I'm not 100% sure why this is implemented the way it is. @jeapostrophe, you fixed a bug in this according to git history, is this change equivalent to the old implementation, and is either buggy?

### DIFF
--- a/rackunit-lib/rackunit/private/format.rkt
+++ b/rackunit-lib/rackunit/private/format.rkt
@@ -1,5 +1,7 @@
 #lang racket/base
-(require racket/match
+(require racket/function
+         racket/list
+         racket/match
          "base.rkt"
          "check-info.rkt"
          "text-ui-util.rkt"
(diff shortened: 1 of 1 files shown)


rackunit-lib/rackunit/private/format.rkt
```diff
@@ -131,27 +133,16 @@
...
+(define (binary-check-this-frame? stack)
+  (ormap check-actual? (takef stack (negate check-name?))))
+
 ;; strip-redundant-parms : (list-of check-info) -> (list-of check-info)
 ;;
 ;; Strip any check-params? is there is an
 ;; actual/expected check-info in the same stack frame.  A
 ;; stack frame is delimited by occurrence of a check-name?
-(define (strip-redundant-params start-stack)
-  (define (binary-check-this-frame? stack)
```
bennn: The new code is "observationally equivalent" to the old, IMO.

No idea why this `(define (binary-check-this-frame? ....` wasn't just `(define binary-check-this-frame? (let loop ....`

jeapostrophe:
I have no idea why this code was the way it was. I find rackunit's implementation to be pretty crazy (which is why I wrote `chk`)

jackfirth:
I'm gonna simplify this to just remove `params` if `actual` is present and ignore all this nonsense about "frames". That's something I can understand the purpose of. Unless someone is parsing the output of `run-tests` I doubt anything will break.

Issue #35, opened 2017-06-14
rackunit/text-ui duplicates a lot of code in rackunit/private/format


jackfirth:
This makes it tricky to implement things like #10. There's also several "nice to have" things in text-ui that could benefit from a more principled API, for instance:

- `run-tests` treats certain info fields as "verbose" fields, specifically the `expression` field. It might be nice to have a `verbose-info` wrapper struct similar to `string-info` that causes the info to not be printed by default and only printed when `run-tests` is run in verbose mode.
- `run-tests` adjusts the source location value in the `location` field to print as a relative directory path. A `location-info` wrapper struct that let users trigger this behavior for other srcloc fields could be useful.
- `run-tests` uses `pretty-print` on certain values. A `pretty-info` wrapper could work, but `run-tests` is a little weird because it pretty prints values in the `expected` and `actual` fields but it *maps* `pretty-print` over the *list* of values in the `params` field. Maybe that behavior should be changed?

A radical approach to this problem is to pull `rackunit/text-ui` out into its own package. I did this as an experiment on my own fork to see what changes need to be made to `text-ui` to remove its dependency on rackunit's internal error formatting code. If that seems like a useful thing to do, I can open a PR.

bennn:
A few thoughts:

#### Implementing Nested Check Info

At first, I don't see how this issue is making #10 difficult. I figure the TODOs for #10 are:
- add a new "check info field" (sorry I don't know if there's a technical term for this)
- design a datatype for this field, made of strings and structs-to-guide formatting
- implement a printer that interprets the datatype and outputs text like #10 suggests

All three of those could be done just in `rackunit/text-ui` right?

I guess the trouble would be if you wanted to re-use a `rackunit/private/format` function at +1 indentation.

Do you want to open issues or list the other things like #10 that you want?

#### More Principled API

All these `-info` structs sound good. I think it'd be great if `rackunit/text-ui` did all its printing in a uniform way, guided by a datatype.

Earlier I was thinking rackunit could re-use structs from `scribble/base`. I don't think so anymore. Except for `nested` I think the goals in Scribble are too different.
Maybe though, it will make sense to move some formatting functions into `racket/format` or some other, general place.

Mapping `pretty-print` sounds like a mistake


#### New Package

How did the moving go? Hopefully it was easy, and the few functions it needed from `rackunit/private/format` could be made public to clients?

I don't think `rackunit/text-ui` needs to be its own package. I think it should:
-  use a public API for building `check-info` values
[...]

PR #38, opened 2017-06-14
Unify `run-tests` and default check handling of `location` field

jackfirth:
This adds a private `location-info` struct alongside `string-info` from #27 to encapsulate rendering a source location with a relative directory path. This obsoletes the logic in both `rackunit/private/format` and `rackunit/text-ui` for handling a check info with key `'location` specially. This is not technically backwards compatible as it changes the observable behavior of `make-check-location`, but it does so in a way that's highly unlikely to affect existing users. See the message of commit 112d4b3 for details.

Works towards closing #35


jeapostrophe:
I think this is good and should be merged. Any other opinions?

rackunit-lib/rackunit/private/check-info.rkt
```diff
@@ -19,14 +20,22 @@
...
+  (cond
+    [(string-info? info-value) (string-info-value info-value)]
+    [(location-info? info-value)
+     (trim-current-directory
+      (location->string (location-info-value info-value)))]
+    [else (with-output-to-string (λ () (write info-value)))]))
+
+(define (trim-current-directory path)
+  (define cd (path->string (current-directory)))
+  (regexp-replace (regexp-quote cd) path ""))
```
bennn: I think `(path->string (find-relative-path (current-directory) path))` will have the same result

rackunit-test/tests/rackunit/text-ui-util-test.rkt
```diff
@@ -1,55 +0,0 @@
...
-;; Commentary:
-#lang racket/base
-
-(require rackunit
-         rackunit/private/text-ui-util)
-
-(provide text-ui-util-tests)
-
-(define text-ui-util-tests
-  (test-suite
```
bennn: Please add these to `private/check-info.rkt`

rackunit-lib/rackunit/private/check-info.rkt
```diff
@@ -19,14 +20,22 @@
...
 (define (info-value->string info-value)
-  (if (string-info? info-value)
-      (string-info-value info-value)
-      (with-output-to-string (λ () (write info-value)))))
+  (cond
+    [(string-info? info-value) (string-info-value info-value)]
+    [(location-info? info-value)
+     (trim-current-directory
+      (location->string (location-info-value info-value)))]
+    [else (with-output-to-string (λ () (write info-value)))]))
```
bennn: Oh hey, this could just be `(~s info-value)` right?
(Maybe `~.s`)

rackunit-lib/rackunit/private/check-info.rkt
jackfirth: Good note, though I'd want to do that in a separate PR.

rackunit-lib/rackunit/private/check-info.rkt
jackfirth: I think `~s` is more appropriate than `~.s` to preserve existing behavior (also chopping off check info seems... weird).

rackunit-test/tests/rackunit/text-ui-util-test.rkt
jackfirth: Done

AlexKnauth:
Is it possible that this change broke something in the typed racket tests that only shows up when I run them in DrRacket?

I'm getting an error that seems to come from [rackunit/private/base, line 27](https://github.com/racket/rackunit/blob/master/rackunit-lib/rackunit/private/base.rkt#L27):
```
(define loc (location-info-value (check-info-value maybe-location)))
```
Where `maybe-location` is `(check-info 'location '("<pkgs>/typed-racket-test/unit-tests/typecheck-tests.rkt" 1436 8 60243 109))`, and `(check-info-value maybe-location)` is `'("<pkgs>/typed-racket-test/unit-tests/typecheck-tests.rkt" 1436 8 60243 109)`, where `location-info-value` is failing because it isn't a `location-value` struct.

I'm only getting this error when I run the TR test file from DrRacket, not when I run it from the command line. (And only when there's a test failure.)

AlexKnauth:
The TR test file uses `with-check-info` instead of going through the `make-check-location`, like this:
```
(with-check-info
  (['location (build-source-location-list (quote-srcloc #,stx))])
  ...)
```
Is that not a supported pattern anymore? Should the documentation for `with-check-info` warn about this? Or should `rackunit/private/base` support both patterns?

bennn:
I think `rackunit` should still support that pattern. Using an actual srcloc for the location field should not error, and rackunit should be able to use the srcloc.

jackfirth:
Meant to respond to this earlier, but alas I did not.

Check infos are weird. They're symbol-value mappings with no namespace controls (which identifiers have via scopes and modules) _and_ no value controls (which hashes and other key-value mappings have via contracts). That's a backwards compatibility nightmare, since people will stick who-knows-what in info values and build frameworks and libraries for reading things from specific info keys without attaching contracts.

I'd prefer if `rackunit` handled check infos with the philosophy "no values are special unless wrapped in a generic `rackunit` info struct like `nested-info`, and no keys are special unless a library uses a specific key with a specialized value wrapper struct like `location-info` to do something extra". Anything else invites inconsistency and makes changing things without breaking things hard.

AlexKnauth:
Is that a reason for `(with-check-info* (list (make-check-location ....)) (λ () ....))` and `(with-check-info (['location ....]) ....)` being different?

PR #40, opened 2017-06-16
Unify `run-tests` and default check handling of pretty-printed infos

jackfirth:
This PR adds a private `pretty-info` struct similar to `string-info` that allows a check info value to be rendered with `pretty-format`. This obsoletes the special cased pretty printing logic of `run-tests`. As a slight formatting tweak, the `'params` info value is now pretty printed as a whole instead of each individual parameter being pretty printed.

This PR builds off of #38 - only the last commit 84c031f is relevant here. The previous commits are duplicated in #38.


bennn:
Cool, I will cherry-pick the last commit after #38 is merged

bennn:
Merged the last commit: https://github.com/racket/rackunit/commit/5570f56e7fa108d81c09339f09b339534e9c051d

PR #41, opened 2017-06-17
Implement check-info with a parameter

### LINKED ISSUE #39: Implement `check-info` stack with a parameter instead of continuation marks
The current internal representation of the check info stack is in terms of continuation marks and a private key. A simpler implementation would be a parameter containing the list of infos which `with-check-info*` manipulates with `parameterize`. Parameters basically work with a private key and continuation marks anyway, so this isn't a significantly different approach.

jackfirth:
This replaces the previous implementation that directly manipulated
continuation marks. After this change, errors raised in tests printed
with run-tests no longer print check info from where the error was
raised. Tests run outside of `run-tests` never did that in the first
place and nobody complained, so I don’t think it’s a big loss compared
to the simpler implementation. A more principled way to do that would
be for `define-check` to wrap unexpected exceptions in an exception that
captures check info.

This also makes it possible in the future to remove the check info sorting logic
in `format.rkt`, since `with-check-info*` now has complete control over the
order check infos are added in.

Closes #39


rackunit-lib/rackunit/private/check-info.rkt
```diff
@@ -11,8 +11,9 @@
                       [value any/c])]
   [struct string-info ([value string?])]
   [info-value->string (-> any/c string?)]
-  [check-info-mark symbol?]
-  [check-info-stack (continuation-mark-set? . -> . (listof check-info?))]
+  [current-check-info (parameter/c (listof check-info?))]
+  ;;[check-info-mark symbol?]
+  ;;[check-info-stack (continuation-mark-set? . -> . (listof check-info?))]
```
bennn: No reason to keep these comments, right?

rackunit-test/tests/rackunit/check-info-test.rkt
```diff
@@ -40,7 +40,7 @@
                              (('a 1)
                               ('b 2)
                               ('c 3))
-                             (check-info-stack (current-continuation-marks)))))
+                              (current-check-info))))
```
bennn: style: indent -1

rackunit-test/tests/rackunit/check-info-test.rkt
```diff
@@ -57,7 +57,7 @@
                               (('d 4)
                                ('e 5)
                                ('f 6))
-                              (check-info-stack (current-continuation-marks))))))
+                               (current-check-info)))))
```
bennn: style: indent -1

rackunit-lib/rackunit/text-ui.rkt
```diff
@@ -106,8 +106,6 @@
          ((error-display-handler) (exn-message exn) exn)))]
     [(test-error? result)
      (let ([exn (test-error-result result)])
-       (when (exn? exn)
-         (textui-display-check-info-stack (check-info-stack (exn-continuation-marks exn))))
```
bennn: This seems like a bad thing to remove. Why not `(textui-display-check-info-stack (current-check-info))` ?

jackfirth: 1) I'm not positive using `(current-check-info)` here would do any good. The thrown exception caused control to leave the body of the check containing the `with-check-info` expression so by this point, `(current-check-info)` should be empty. This previous implementation worked because the exception stored the continuation marks from when it was raised and check info could be retrieved from those marks.

2) I removed this because it's a strange asymmetry between `run-tests` and how checks normally run. The default error handler for checks doesn't do this, so it only happens when using `run-tests` with test suites. If it's worth preserving this feature, then we should keep the existing continuation mark based implementation and document why we're not using a parameter.

bennn: ok, now I agree about removing these 2 lines

rackunit-test/tests/rackunit/check-info-test.rkt
jackfirth: DrRacket doesn't seem to like the indentation of this file in general, but that can be fixed another time.

rackunit-lib/rackunit/private/check-info.rkt
jackfirth: Absolutely right!

rackunit-test/tests/rackunit/check-info-test.rkt
jackfirth: Fixed

jackfirth:
Ready for review round 2, pending the fix in #43

jackfirth:
Approved offline by @bennn, squashing and merging.

PR #45, opened 2017-06-26
Add a verbose-info type

jackfirth:
This lets RackUnit mark some info types as “verbose only”, and they’ll
only show up when run-tests is used in ‘verbose mode. Normal checks
evaluated in DrRacket won’t show verbose info.

The only info type marked as verbose is `expression` which is typically
redundant in DrRacket, since the check failure message is supposed to
have an exception srcloc pointing to the failure expression. Checks run
in CI or from the command line won’t show the expression, but the error
message does already include the file and line number.

This does not expose the verbose-info struct to RackUnit users.
Instead, it is used internally to remove the now-redundant verbose-info
logic in rackunit/text-ui. With this change, rackunit/text-ui and
normal check evaluation are at “feature parity” in terms of how they
treat check failures, and the run-tests implementation can be stripped
down and the monad logic removed (see #35).

### DIFF
--- a/rackunit-lib/rackunit/private/format.rkt
+++ b/rackunit-lib/rackunit/private/format.rkt
@@ -4,9 +4,7 @@
          "base.rkt"
          "check-info.rkt")
 
-(provide display-check-info-name-value
-         display-check-info
-         display-check-info-stack
+(provide display-check-info-stack
          display-test-name
          display-exn
 
@@ -14,12 +12,7 @@
          display-failure
          display-error
 
-         display-test-failure/error
-         strip-redundant-params
-         check-info-stack-max-name-width
-
-         display-verbose-check-info
-         sort-stack)
+         display-test-failure/error)
 
 ;; name-width : integer
 ;;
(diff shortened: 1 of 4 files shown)


rackunit-lib/rackunit/private/format.rkt
```diff
@@ -66,40 +59,23 @@
...
   (apply max 0
          (map check-info-name-width check-info-stack)))
 
 ;; display-check-info-stack : (listof check-info) -> void
-(define (display-check-info-stack check-info-stack)
-  (display-verbose-check-info-stack
-   (strip-redundant-params check-info-stack)))
+(define (display-check-info-stack check-info-stack #:verbose? [verbose? #f])
+  (define stack
+    (if verbose? check-info-stack (strip-redundant-params check-info-stack)))
```
bennn: Can `strip-redundant-params` also string `verbose-info?`

If not, I propose you add a `filter` in the else branch here (and remove `should-show?` below)

jackfirth: Done. Renamed `strip-redundant-params` to `simplify-params` as well.

PR #46, opened 2017-06-26
Simplify trim-current-directory implementation

jackfirth:


rackunit-lib/rackunit/private/check-info.rkt
```diff
@@ -42,8 +43,7 @@
     [else (~s info-value)]))
 
 (define (trim-current-directory path)
-  (define cd (path->string (current-directory)))
-  (regexp-replace (regexp-quote cd) path ""))
+  (path->string (find-relative-path (current-directory) path)))
```
bennn: the travis failures go away if you add `#:more-than-root? #t`

jackfirth:
It looks like `find-relative-path` has a bug. The following works:

```racket
(find-relative-path "/foo/bar" "/foo/baz")
```

But this throws an internal contract error claiming `path-system-type` expected `path-for-some-system?` but was given `"/foo/bar"`:

```racket
(find-relative-path "/foo/bar" "/foo/baz" #:more-than-root? #t)
```

rackunit-lib/rackunit/private/check-info.rkt
```diff
@@ -42,8 +43,7 @@
     [else (~s info-value)]))
 
 (define (trim-current-directory path)
-  (define cd (path->string (current-directory)))
-  (regexp-replace (regexp-quote cd) path ""))
+  (path->string (find-relative-path (current-directory) path #:more-than-root? #t)))
```
bennn: The result isn't always a path.

With that change, all tests pass for me after https://github.com/racket/racket/commit/04c1c15d899cb017b4a3ad4ce0088799c6c49364

jackfirth:
I think I'll hold off on this change until after the next release. It's a tiny improvement and to me it's not worth the extra trouble of using the right base Racket version while developing RackUnit (I don't have nightly installed, I'm on 6.9).

jackfirth:
Seems I deleted the fork this was originally in. Reopened as #72.

PR #48, opened 2017-06-27
Replace rackunit/text-ui with a simpler implementation

### LINKED ISSUE #35: rackunit/text-ui duplicates a lot of code in rackunit/private/format

jackfirth:
Closes #35 

This throws the old implementation out, which was implemented with a
strange ad-hoc dynamic (and impure!) monad-like abstraction backed with
a hash map and lots of lambdas. Now a simple struct is used to manage
the test fold state. Additionally, all duplicated logic in
`rackunit/private/format` is removed; `rackunit/text-ui` now contains no
knowledge of the default output format of check messages.


rackunit-lib/rackunit/private/format.rkt
```diff
@@ -1,31 +1,38 @@
...
+  (cond [(test-failure? res) (display-err (test-failure-result res))]
+        [(test-error? res) (display-err (test-error-result res))]
+        [else (void)]))
+
+(define (combine-names test-name suite-names)
+  (define any-names? (or test-name (not (empty? suite-names))))
+  (and any-names?
+       (string-join (snoc (or test-name "Unnamed test") suite-names) " > ")))
+
+(define (snoc v vs) (append vs (list v)))
```
bennn: Instead of `snoc`, use the `#:after-last` option to `string-join` ?

rackunit-lib/rackunit/text-ui.rkt
```diff
@@ -28,139 +28,80 @@
...
-                      (apply (display-counter) args))))])))
-
-;; run-tests : test [(U 'quiet 'normal 'verbose)] -> integer
+
+;; As we fold over the test results, there are two pieces of state we must
+;; manage: a counter of failures, errors, and passes, and a stack of names
+;; indicating where we are in the tree of test suites and cases. The former is
+;; used for printing a summary message after running all tests, while the latter
+;; is used in each test to print what cases / suites a test is in.
+(struct fold-state (counter names) #:transparent)
```
bennn: why not fields `[successes failures errors names]` ?

A hashtable seems like too much data structure

rackunit-lib/rackunit/text-ui.rkt
```diff
@@ -28,139 +28,80 @@
...
+  ;; we install a new custodian to ensure resources handled improperly by tests
+  ;; such as threads, tcp listeners, etc. don't interfere with other tests or
+  ;; the runner itself
+  (define counter (fold-state-counter (call/new-custodian fold-tests)))
+  (unless quiet?
+    (display-counter counter))
+  (num-unsuccessful counter))
+
+(define (call/new-custodian thnk)
+  (parameterize ([current-custodian (make-custodian)]) (thnk)))
```
bennn: Need to call `custodian-shutdown-all` ?

rackunit-lib/rackunit/text-ui.rkt
jackfirth: Picking the right field to update dynamically in `increment-counter` gets messy and complicated when using struct fields instead of a hash table. If `struct` defined functional setters / updaters it'd probably be the better option.

rackunit-lib/rackunit/text-ui.rkt
jackfirth: The old code didn't shut down the custodian so I don't think it's needed. It's enough to isolate the resources in a separate custodian, the test runner doesn't need to enforce that they're cleaned up in a specific timeframe.

rackunit-lib/rackunit/private/format.rkt
jackfirth: `#:after-last` doesn't insert the separator, so I think adding the test name to the list directly makes more sense.

PR #50, opened 2017-06-27
Fix missing calls to error-display-handler for checks without failure messages

### LINKED ISSUE #42: Check failures without messages don't show the error location in DrRacket
When `fail-check` is called without a message (the default behavior for most checks) the default check failure printer skips over the message section, but in doing so it omits the call to `error-display-handler` that prints the exception symbol in DrRacket and highlights the check expression in red. This is caused by a bug in #25.

A possibly more appropriate place for the exception symbol would be right before the "FAILURE" / "ERROR" text in the check output. This text always exists regardless of whether the check has a message or not.

jackfirth:
Closes #42 

Not using `error-display-handler` causes DrRacket to lose the clickable “stop signs” in check failures that points to the check’s location; this adds back the missing calls. Also, cleans up the failure printing logic a bit since that can now be safely done without messing up `rackunit/text-ui`.

cc @AlexKnauth

### DIFF
--- a/rackunit-test/tests/rackunit/standalone.rkt
+++ b/rackunit-test/tests/rackunit/standalone.rkt
@@ -41,8 +41,8 @@
            #"\
 --------------------
 ERROR
-Outta here!
 
+Outta here!
 --------------------
 --------------------
 FAILURE
@@ -58,14 +58,14 @@ message:    0.0
            #"\
 --------------------
 ERROR
-First Outta here!
 
+First Outta here!
 --------------------
 --------------------
 error
 ERROR
-Second Outta here!
 
+Second Outta here!
 --------------------
 --------------------
 FAILURE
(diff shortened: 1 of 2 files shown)


rackunit-test/tests/rackunit/standalone.rkt
```diff
@@ -41,8 +41,8 @@
            #"\
 --------------------
 ERROR
-Outta here!
 
+Outta here!
```
bennn: Why not keep this the same, and remove the `newline` on L114 of `format.rkt` ?

jackfirth: I think it looks kinda weird; the extra newline after the message doesn't seem to serve a purpose. Changing it to print this way allows common code to handle printing failures and errors.

PR #51, opened 2017-07-01
Remove check info stack sorting

jackfirth:
The new parameter representation of infos ensures that the built-in
`define-check` infos appear in order before any user-added infos. The
one exception is that the `expression` info now appears before the
`actual` or `expected` infos, but since it’s only present in verbose
mode I don’t think that’s much of an issue.

### DIFF
--- a/rackunit-lib/rackunit/private/format.rkt
+++ b/rackunit-lib/rackunit/private/format.rkt
@@ -73,7 +73,7 @@
   (define stack
     (if verbose? check-info-stack (simplify-params check-info-stack)))
   (define max-name-width (check-info-stack-max-name-width stack))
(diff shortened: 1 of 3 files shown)


bennn:
Are there tests for the sorting?

Can you add a test that shows `expression` comes before `actual`/`expected`?

jackfirth:
I added two tests. One verifies that `define-check` adds infos in a certain order, the other verifies that infos added inside the body of `define-check` occur after infos added by `define-check`. Together, these verify that `actual` / `expected` occur after `expression` without deliberately failing a use of `check-equal?`. I wanted to avoid that since it requires using a separate namespace so the failure `check-equal?` logs is ignored.

PR #52, opened 2017-07-01
Add nested-info

### LINKED ISSUE #10: Nested check info?

jackfirth:
Closes #10, depends on #51 

This PR adds a `(nested-info info ...)` struct that allows custom checks to print an indented list of check infos under a single info key. The following check:

```racket
(define-check (foo)
  (define nested
    (nested-info (make-check-info 'bar 'bar)
                 (make-check-info 'baz 'baz)))
  (with-check-info (['foo nested])
    (fail-check)))
```

Raises a check failure with output like this:

```
--------------------
FAILURE
name:       foo
location:   unsaved-editor:10:0
params:     '()
foo:
  bar:        bar
  baz:        baz
--------------------
```

### DIFF
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -269,7 +269,8 @@ the @racket[with-check-info*] function, and the
 @defstruct[check-info ([name symbol?] [value any]) #:transparent]{
  A check-info structure stores information associated with the context of the
  execution of a check. The @racket[value] is written in a check failure message
- using @racket[write] unless it is a @racket[string-info] value.
+ using @racket[write] unless it is a @racket[string-info] value or a
+ @racket[nested-info] value.
  @history[#:changed "1.6" "Changed from opaque to transparent"]}
 
 @defstruct*[string-info ([value string?])]{
@@ -278,10 +279,29 @@ the @racket[with-check-info*] function, and the
  quotes. Used to print messages in check infos instead of writing values.
  @(interaction
    #:eval rackunit-eval
-   (with-check-info (['value "hello world"]
-                     ['message (string-info "hello world")])
-     (check = 1 2)))
+   (define-check (string-info-check)
+     (with-check-info (['value "hello world"]
+                       ['message (string-info "hello world")])
+       (fail-check)))
+   (string-info-check))
  @history[#:added "1.2"]}
+
+@deftogether[
+ ((defproc (nested-info [info check-info?] ...) nested-info?)
+  (defproc (nested-info? [v any/c]) boolean?)
+  (defproc (nested-info-values [nested nested-info?]) (listof check-info?)))]{
+ A special wrapper around a list of infos for use as a @racket[check-info]
+ value. A check info whose value is a nested info is displayed as an indented
+ sub-list of infos. Nested infos can be placed inside nested infos, yielding
+ greater indentation.
+ @(interaction
+   #:eval rackunit-eval
+   (define-check (nested-info-check)
+     (define nested
+       (nested-info (make-check-info 'foo "foo") (make-check-info 'bar "bar")))
+     (with-check-info (['nested nested]) (fail-check)))
+   (nested-info-check))
+ @history[#:added "1.7"]}
 
 The are several predefined functions that create check
 information structures with predefined names.  This avoids
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -273,7 +273,7 @@ the @racket[with-check-info*] function, and the
  @racket[nested-info] value.
  @history[#:changed "1.6" "Changed from opaque to transparent"]}
 
-@defstruct*[string-info ([value string?])]{
+@defstruct*[string-info ([value string?]) #:transparent]{
  A special wrapper around a string for use as a @racket[check-info] value. When
  displayed in a check failure message, @racket[value] is displayed without
  quotes. Used to print messages in check infos instead of writing values.
@@ -286,10 +286,7 @@ the @racket[with-check-info*] function, and the
    (string-info-check))
  @history[#:added "1.2"]}
 
-@deftogether[
- ((defproc (nested-info [info check-info?] ...) nested-info?)
-  (defproc (nested-info? [v any/c]) boolean?)
-  (defproc (nested-info-values [nested nested-info?]) (listof check-info?)))]{
+@defstruct*[nested-info ([values (listof check-info?)]) #:transparent]{
  A special wrapper around a list of infos for use as a @racket[check-info]
  value. A check info whose value is a nested info is displayed as an indented
  sub-list of infos. Nested infos can be placed inside nested infos, yielding
@@ -297,9 +294,9 @@ the @racket[with-check-info*] function, and the
  @(interaction
    #:eval rackunit-eval
    (define-check (nested-info-check)
-     (define nested
-       (nested-info (make-check-info 'foo "foo") (make-check-info 'bar "bar")))
-     (with-check-info (['nested nested]) (fail-check)))
+     (define infos
+       (list (make-check-info 'foo "foo") (make-check-info 'bar "bar")))
+     (with-check-info (['nested (nested-info infos)]) (fail-check)))
    (nested-info-check))
  @history[#:added "1.7"]}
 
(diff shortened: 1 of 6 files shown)


rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -269,19 +269,39 @@ the @racket[with-check-info*] function, and the
 @defstruct[check-info ([name symbol?] [value any])]{
  A check-info structure stores information associated with the context of the
  execution of a check. The @racket[value] is written in a check failure message
- using @racket[write] unless it is a @racket[string-info] value.}
+ using @racket[write] unless it is a @racket[string-info] value or a
+ @racket[nested-info] value.}
```
bennn: This doc sounds okay now, but it's going to be awkward if there are more formatting structs.

It'd be better to have a predicate for the formatting structs & reference that function here ("... using `@racket[write]` unless it is a ???.")

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -269,19 +269,39 @@ the @racket[with-check-info*] function, and the
...
-                     ['message (string-info "hello world")])
-     (check = 1 2)))
+   (define-check (string-info-check)
+     (with-check-info (['value "hello world"]
+                       ['message (string-info "hello world")])
+       (fail-check)))
+   (string-info-check))
  @history[#:added "1.2"]}
 
+@deftogether[
```
bennn: I think this should just be a `defstruct*`. Let them write `(apply nested-info v ....)`.

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -269,19 +269,39 @@ the @racket[with-check-info*] function, and the
...
+ ((defproc (nested-info [info check-info?] ...) nested-info?)
+  (defproc (nested-info? [v any/c]) boolean?)
+  (defproc (nested-info-values [nested nested-info?]) (listof check-info?)))]{
```
bennn: `(listof check-info?)` isn't right, right? Rules out `(list (list (check-info ....)))`.

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -269,19 +269,39 @@ the @racket[with-check-info*] function, and the
...
+ A special wrapper around a list of infos for use as a @racket[check-info]
+ value. A check info whose value is a nested info is displayed as an indented
+ sub-list of infos. Nested infos can be placed inside nested infos, yielding
```
bennn: I'd prefer "sequence" or "newline-separated sequence" over "sub-list".

Sub-list makes me think Racket lists.

rackunit-lib/rackunit/private/check-info.rkt
```diff
@@ -32,6 +35,10 @@
 (struct location-info (value) #:transparent)
 (struct pretty-info (value) #:transparent)
 (struct verbose-info (value) #:transparent)
+(struct nested-info (values)
+  #:transparent #:omit-define-syntaxes #:constructor-name make-nested-info)
+
+(define (nested-info . vs) (make-nested-info vs))
```
bennn: Same comment as above: I think you should just remove this.

rackunit-lib/rackunit/private/format.rkt
```diff
@@ -33,50 +35,48 @@
...
-   [(m . < . n)
-    (string-append s (make-string (- n m) #\space))]
-   [else
-    (substring s 0 n)]))
+(define (string-padding str desired-len)
+  (make-string (max (- desired-len (string-length str)) 0) #\space))
+
+(define (string-indent str amnt)
+  (define pad (make-string amnt #\space))
+  (string-append pad (string-replace str "\n" (string-append "\n" pad))))
```
bennn: Why do newlines need to be replaced?

rackunit-lib/rackunit/private/format.rkt
```diff
@@ -33,50 +35,48 @@
...
-  (match a-check-info
-    [(struct check-info (name value))
-     (display-check-info-name-value max-name-width name value)]))
-
-(define (check-info-stack-max-name-width check-info-stack)
-  (apply max 0
-         (map check-info-name-width check-info-stack)))
+(define (check-info-stack-name-width check-info-stack)
+  (define widths (map check-info-name-width check-info-stack))
+  (apply max 0 widths))
```
bennn: Don't see why you need `0` here, since widths come from `string-length`.

rackunit-test/tests/rackunit/nested-info-test.rkt
```diff
@@ -0,0 +1,44 @@
...
+       (make-check-info
+        'nested (nested-info
+                 (make-check-info
+                  'double-nested (nested-info (make-check-info 'foo 1)
+                                              (make-check-info 'bar 2)))))))
+    (define expected-str "nested:
+  double-nested:
+    foo:            1
+    bar:            2
+")
```
bennn: Can you add tests for:
- 2 or more "double-nested"s
- extra newlines in an `expected-str` (to answer my question from above, why do newlines need to be padded)

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: Agreed, though I'd like to wait until we've got a third public documented one of these before making that change.

rackunit-lib/rackunit/private/format.rkt
jackfirth: `string-indent` takes a string and indents every line of that string to the right, not just the first line. So each newline needs to be replaced by a newline with padding, so that this:

```
foo
bar
baz
```

becomes this:

```
    foo
    bar
    baz
```

Without the newline replacement, you'd get this:

```
    foo
bar
baz
```

rackunit-lib/rackunit/private/format.rkt
jackfirth: It's in case `check-info-stack` is empty; `max` throws an exception if given zero arguments.

rackunit-test/tests/rackunit/nested-info-test.rkt
jackfirth: Do you mean two nested infos, or one nested info containing two nested infos?

bennn: one nested containing 2 nesteds

rackunit-lib/rackunit/private/check-info.rkt
jackfirth: Done

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: `(listof check-info?)` is right, it's the field in `nested-info`.

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: Done

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: Done

rackunit-test/tests/rackunit/nested-info-test.rkt
jackfirth: Done

jackfirth:
I think I've addressed all your comments @bennn

PR #57, opened 2017-07-04
Simplify define-check implementation

jackfirth:


rackunit-lib/rackunit/private/check.rkt
```diff
@@ -78,71 +81,51 @@
...
+                  #:location (syntax->location #'loc)
+                  #:expression 'chk
+                  args)))]))
+
+(define-simple-macro (define-check (name:id formal:id ...) body:expr ...)
+  (begin
+    (define-check-func (check-impl formal ...) #:public-name name body ...)
+    (define-syntax (name stx)
+      (syntax-parse stx
+        [(_ arg (... ...))
```
bennn: could this just be `(_ . arg)` ?

rackunit-lib/rackunit/private/check.rkt
```diff
@@ -78,71 +81,51 @@
...
+              #:expression '(chk arg ...)))]))
+
+(define-syntax (id-check stx)
+  (syntax-parse stx
+    [(_ chk:id)
+     (with-syntax ([loc (datum->syntax #f 'loc stx)])
+       #`(lambda args
+           (apply chk
+                  #:location (syntax->location #'loc)
+                  #:expression 'chk
```
bennn: Could this be `` `(chk . ,args)`` ?

jackfirth: No, that would capture the runtime values of args rather than the expression that produced them. The runtime values are already captured in the `params` info.

rackunit-lib/rackunit/private/check.rkt
jackfirth: Yup, thanks!

jackfirth:
The intent is to clearly separate the macro into four pieces: the function definition, the location-capturing function application, the location-capturing non-application expression, and the high-level macro that stitches the three together. Long macro implementations give me the heebie jeebies.

rackunit-lib/rackunit/private/check.rkt
bennn: Ok.
Should this be `(chk (gensym formal) ...)` like it was before (more or less)?

jackfirth: Personally I don't think so, because it obscures that the check was used in a higher-order way without adding any useful information.

jackfirth:
I see what you mean. I removed `app-check` and `id-check*` and it's much simpler.

Issue #55, opened 2017-07-02
Check contracts


jackfirth:
Since custom checks aren't functions, they can't be protected by `contract-out` without disabling their location-capturing capabilities. A `check-out` provide form that allows contracts to be added could work like this:

```racket
(provide
  (check-out check-foo (check-> foo?)))

(define-check (check-foo foo) ...)
```

A check that has its contract violated should raise an error rather than causing a check failure. The `check->` combinator is used instead of a normal `->` combinator because 1) checks always return `void?`, 2) checks have extra "secret" arguments (`#:location` and `#:expression` specifically) passed in by their macro expansion, and 3) checks don't support user-defined keyword, optional, or rest arguments.

bennn:
Hmm, `check-out` seems like a bad way to fix because it's new & very special purpose syntax.

What about, instead, `define-check/contract`? 

(Also since `check->` is nothing like `->` I think it should be named `check/c` ... but that's besides the point.)

jackfirth:
`contract-out` and `define/contract` use different parties for blaming: `contract-out` uses a boundary between the exporting and importing modules, but `define/contract` uses a boundary between the function itself and the rest of the module it's defined in. `define-check/contract` would definitely be useful, but it could only work in the way `define/contract` does where there's a boundary between the check and it's surrounding module. I think we'd need some sort of `provide` form to blame in terms of client modules.

Also, +1 to `check/c`.

bennn:
Good point about the boundaries.

Well I think the best fix would be changing `contract-out` to handle macros.
I'm not sure how to do that, so +1 to `check-out`.

PR #59, opened 2017-07-06
WIP - Add meta checks

### LINKED ISSUE #14: Testing a check with a check
Hi,

I'm trying to write a test for a custom check but I can't seem to figure it out. The check under test shows up as a failed test in the `raco test` output. Is there a way to do this? If so, what am I doing wrong?

My check

```
(define-check (check-mock-called-with? args mock)
  (let ([result (mock-called-with? args mock)])
    (with-check-info
      [('expected args) ('actual (mock-calls mock))]
      (unless result
        (fail-check "Actual arguments did not match expected")))))
```

The test for the check

```
  (test-case
    "check-info is added by check-mock-called-with? on failure"
    (check-exn
      (λ (e) (and (has-info? e 'expected) (has-info? e 'actual)))
      (thunk
        (parameterize ([current-check-handler raise])
          (check-mock-called-with? '(1) (void-mock)))))))
```

Running the test yields 1/2 failed tests. The succeeded test is the outer one (which is good) but the inner one shows up as a failure, which it shouldn't because that's part of the test. I expected the inner test to not register in the raco test output because its exception is raised by the current check handler instead of printed.

More info: https://github.com/jackfirth/racket-mock/pull/20

### LINKED ISSUE #47: Proposed meta-check API
As mentioned in #14, it's very difficult to test custom RackUnit checks. This issue proposes some additional checks for RackUnit to improve the situation

### The checks

- `(check-fail (lambda () (check-equal? 1 2)))` - asserts that the given thunk raises a single check failure
- `(check-fail/info (make-actual-info 1) (lambda () (check-equal? 1 2)))` - asserts that the given thunk raises a single check failure containing the given check info
- `(check-fail/message #rx"foo" (lambda () (check-equal? 1 2)))` - asserts that the given thunk raises a single check failure whose `fail-check` message matches the given regexp (a predicate could also be provided, to match the behavior of `check-exn`)

These checks would need to ensure the checks to be tested don't increment the test failure counter, as well as ensuring they don't print their typical output. This will likely require some modification of `define-check` to allow selectively disabling the test failure counter behavior.

### Alternatives considered

- Keyword arguments in `check-fail` for asserting the info and message. This might be nicer to read, but then one check expression is asserting multiple things. This conflicts with the typical philosophy of "one check = one assertion". Plus, keyword arguments aren't supported in custom checks (see #17).
- Taking expressions instead of thunks. Existing checks like `check-exn` take thunks, and `define-check` doesn't allow delaying evaluation.
- There's no point to a `check-pass` form - just use the check directly.
- No check is provided for making assertions about the *entire* info stack instead of a single info value, or for checking an info value matches a particular predicate. The info stack is intended to be read by humans, not programmatically manipulated, so its structure is not something that should be rigorously enforced.

jackfirth:
### DO NOT MERGE - Work in progress

Closes #14, closes #47 

This PR adds the following checks:

- `(check-fail <pred-or-exn> <thunk>)` - asserts that `<thunk>` evaluates a failing check and that the check failure exception satisfies `<pred-or-exn>` in the same way as `check-exn`
- `(check-fail* <thunk>)` - like `check-fail`, but without caring about the specifics of the thrown failure
- `(check-fail/info <check-info> <thunk>)` - asserts that `<thunk>` evaluates a failing check and that the check failure exception contains an info that is `equal?` to `<check-info>`
- `(check-error <pred-or-exn> <thunk>)` - asserts that `<thunk>` evaluates a check that raises an error instead of failing and that the raised value satisfies `<pred-or-exn>`

I went with `check-fail/info` instead of adding an info case to `check-fail` to avoid confusion over whether `check-error` has an info-accepting variant. The existence of `check-fail/info` and nonexistence of `check-error/info` is a clear signal in documentation search results and playing at the REPL that wouldn't be present if `check-fail/info` was merged with `check-fail`.

The only addition from the proposed API is `check-error`, which I realized I'd need to test to the contract failure cases of these checks.

### TODO

- [ ] Add contract failure tests
- [ ] Add documentation
- [ ] Bump version


AlexKnauth:
Why is check-error called `check-error`?

jackfirth:
Symmetry with the output of checks, since check failures print `"FAILURE"` while check errors print `"ERROR"`. If you have a suggestion for a less ambiguous name, I'm all ears.

bennn:
How does `check-error` differ from `check-exn`?

I tried a quick example and `check-exn` works how I'd expect:
```
#lang racket
(module+ test
  (require rackunit)
  (check-exn exn:fail:contract?
    (lambda () (check-true (+ "asdf" 2)))))
```

```
$ raco test test.rkt
raco test: (submod "test.rkt" test)
1 test passed
```

Reminds me though, it would be good to have a `check-exn` helper to test "throws a contract error that blames a certain party"

jackfirth:
That example works because `(+ "asdf" 2")` is evaluated before being passed as an argument to `check-true`, so the error isn't thrown from within the body of `check-true`. Try this example:

```racket
(check-exn (lambda (v) (equal? v 'oops))
  (check-pred (lambda (v) (raise 'oops)) 'foo))
```

rackunit-lib/rackunit/private/meta.rkt
```diff
@@ -0,0 +1,120 @@
...
+
+
+(define-check (check-fail pred-or-msg chk-thnk)
+  (contract-pred-or-msg! 'check-fail pred-or-msg)
+  (contract-thunk! 'check-fail chk-thnk)
+  (define failure (check-raise-value chk-thnk))
+  (with-expected pred-or-msg
+    (assert-failure failure)
+    (assert-check-failure failure)
+    (if (procedure? pred-or-msg)
```
bennn: I think `(predicate/c pred-or-msg)` will do the right thing (and also check `(procedure-arity-includes? pred-or-msg 1)`)

(or use `contract-out`)

rackunit-lib/rackunit/private/meta.rkt
```diff
@@ -0,0 +1,120 @@
...
+        (unless (pred-or-msg failure)
+          (fail-check "Check failure didn't pass predicate"))
+        (unless (regexp-match? pred-or-msg (exn-message failure))
+          (fail-check "Check failure message didn't match regexp")))))
```
bennn: These `fail-check` should print `failure` somehow. (Do they?)

I'd prefer they mirror `check-exn`, and also say `"Wrong failure raised"` in both cases

rackunit-lib/rackunit/private/meta.rkt
```diff
@@ -0,0 +1,120 @@
...
+    (unless (for/or ([name (in-list info-names)])
+              (equal? name (check-info-name expected-info)))
+      (with-check-info (['actual-info-names info-names])
+        (fail-check "Check failure did not contain the expected info")))
+    (for ([info (in-list infos)])
+      (when (equal? (check-info-name info) (check-info-name expected-info))
+        (unless (equal? info expected-info)
+          (with-actual info
+            (fail-check
+             "Check failure contained an unexpected info value")))))))
```
bennn: This stops at the first unexpected info, right?

I think it'd be better to print all the unexpecteds.

rackunit-lib/rackunit/private/meta.rkt
```diff
@@ -0,0 +1,120 @@
...
+
+(define-check (check-error pred-or-msg chk-thnk)
+  (contract-pred-or-msg! 'check-error pred-or-msg)
+  (contract-thunk! 'check-error chk-thnk)
+  (define failure (check-raise-value chk-thnk))
+  (with-expected pred-or-msg
+    (assert-failure failure)
+    (assert-not-check-failure failure)
+    (cond
+      [(procedure? pred-or-msg)
```
bennn: `(predicate/c pred-or-msg)`

rackunit-lib/rackunit/private/meta.rkt
```diff
@@ -0,0 +1,120 @@
...
+
+;; Shorthands for adding infos
+
+(define-simple-macro (with-actual act:expr body:expr ...)
+  (with-check-info* (list (make-check-actual act)) (λ () body ...)))
+
+(define-simple-macro (with-expected exp:expr body:expr ...)
+  (with-check-info* (list (make-check-expected exp)) (λ () body ...)))
+
+;; Pseudo-contract helpers, to be replaced with real check contracts eventually
```
bennn: eventually = this pull request?

rackunit-lib/rackunit/private/meta.rkt
```diff
@@ -0,0 +1,120 @@
...
+
+(define (contract-pred-or-msg! name pred-or-msg)
+  (unless (or (procedure? pred-or-msg) (regexp? pred-or-msg))
```
bennn: `predicate/c` 
(next line too)

rackunit-lib/rackunit/private/meta.rkt
```diff
@@ -0,0 +1,120 @@
...
+(define (check-raise-value chk-thnk)
+  ;; To fully isolate the evaluation of a check inside another check,
+  ;; we have to ensure that 1) the inner check raises its failure normally
+  ;; instead of writing to stdout / stderr, 2) the inner check doesn't log
+  ;; any pass or fail information to rackunit/log, and 3) the inner check's info
+  ;; stack is independent of the outer check's info stack.
+  (parameterize ([current-check-handler raise]
+                 [test-log-enabled? #f]
+                 [current-check-info (list)])
+    (with-handlers ([(λ (e) (not (exn:break? e))) values]) (chk-thnk) #f)))
```
bennn: use `negate` ? 

I don't really care but `racket/function`'s already imported

rackunit-lib/rackunit/private/meta.rkt
jackfirth: By "real check contracts", I mean #55. I don't plan on implementing those in this PR.

rackunit-lib/rackunit/private/meta.rkt
jackfirth: I'd rather not add the complexity of real higher order contracts here. Additionally, the function used as a predicate can return non-boolean values which `predicate/c` doesn't allow.

rackunit-lib/rackunit/private/meta.rkt
jackfirth: That's a higher order contract, so I don't think it would work here.

rackunit-lib/rackunit/private/meta.rkt
jackfirth: I hadn't thought about if the info stack contains multiple infos with the expected name but different values. This needs to be reworked to deal with that.

rackunit-lib/rackunit/private/meta.rkt
bennn: ok. then `(and (procedure? pred-or-msg) (procedure-arity-includes? pred-or-msg 1))`
?

jackfirth: Done

rackunit-lib/rackunit/private/meta.rkt
jackfirth: Done

rackunit-lib/rackunit/private/meta.rkt
jackfirth: Done. They now print the message and the failure value itself, just like `check-exn`.

jackfirth:
@bennn Added docs and addressed all your comments (I think), this PR is ready for a final review.

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -454,4 +454,86 @@ Raises an @racket[exn:test:check] with the contents of the check
...
+         void?]{
+ Checks that @racket[thunk] evaluates a failing check and that
+ @racket[fail-exn-predicate], it it's a function, returns a true value when
+ given the check failure exception. If @racket[fail-exn-predicate] is a regexp,
+ instead checks that the regexp matches the check failure exception's message.
+ Note that a check failure exception's message is the message given to
+ @racket[fail-check], not the optional @racket[message] argument that all checks
+ accept. See also @racket[check-exn] and @racket[check-error].
+
+ @(interaction
```
bennn: Should get the same output with `@examples[#:label #f ....`

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -454,4 +454,86 @@ Raises an @racket[exn:test:check] with the contents of the check
...
+
+ @history[#:added "1.8"]}
+
+@defproc[(check-error [fail-exn-predicate
+                       (or/c (-> exn:test:check? any/c) regexp?)]
+                      [thunk (-> any)]
+                      [message string? ""])
+         void?]{
+ Checks that @racket[thunk] evaluates a check that raises an error value instead
+ of passing or failing, and checks that the raised value, if it it's a function,
```
bennn: `if it` => `if`

Also "checks that the raised value" should refer to `fail-exn-predicate`.

(The sentence is a little long & awkward ... maybe start a new sentence after "passing or failing"? or end it with "checks that the raised value satisfies the predicate. Satisfies means ....")

rackunit-lib/rackunit/private/meta.rkt
bennn: can you add a `procedure-arity-includes?` check here?

rackunit-lib/rackunit/private/meta.rkt
bennn: same here, add `procedure-arity-includes?`

AlexKnauth:
Would it be better if `check-error` were renamed to either `check-check-error` or `check-fail-error`?

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -454,4 +454,86 @@ Raises an @racket[exn:test:check] with the contents of the check
...
+behavior of other checks; they allow verifying checks pass and fail when
+expected or that checks add certain information to the check information stack.
+
+@defproc[(check-fail [fail-exn-predicate
+                      (or/c (-> exn:test:check? any/c) regexp?)]
+                     [thunk (-> any)]
+                     [message string? ""])
+         void?]{
+ Checks that @racket[thunk] evaluates a failing check and that
+ @racket[fail-exn-predicate], it it's a function, returns a true value when
```
jackfirth: Done.

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: Done. Didn't add `#:label #f` because using the label seems appropriate. They are examples, after all.

jackfirth:
@AlexKnauth How about `check-fail/error`? According to scheme pronunciation rules, that's "check fail with error" which seems accurate.

jackfirth:
@bennn For some reason I can't respond directly to your comments about `procedure-arity-includes?`. The arity check is already present in `contract-pred-or-msg!` so I don't think it needs to be tested in the `cond` expression.

AlexKnauth:
(Re: `check-fail/error`) That makes sense to me.

rmculpepper:
I think this should be a new module, maybe `rackunit/meta`. That would make it clearer that these are not general-purpose checks, but rather specialized for testing other checks.

I'd prefer if `check-fail`, `check-fail*`, and `check-fail/info` were combined into one form. (I disagree with your rationale for keeping them separate.) I propose:

    ;; check-fail : (Treeof FailurePred) (-> Any) -> Void
    ;; where FailurePred is one of
    ;; - (exn:test:check -> Boolean)  -- exn must satisfy predicate
    ;; - Regexp                       -- exn message must match
    ;; - Check-Info                    -- exn must contain check-info

More descriptive, less ambiguous names would be `check-check-fail` and `check-check-error`... but I could understand not liking them. `check-fail/exn` is not quite right, since we consider failures distinct from errors. But `check-error` is bad, because it's too similar to `check-exn`. (I believe there's a library somewhere that defines `check-error` as a macro that wraps the expression to check as a thunk.)

jackfirth:
@rmculpepper Thanks for taking a look at this! Here's my thoughts:

A `rackunit/meta` module is a good idea, but why stop there? Should there be `rackunit/base`, `rackunit/check`, `rackunit/suite`, or other modules for splitting up the contents of `rackunit`? I'd find it strange to only have `rackunit/meta` provided from a special module while nothing else in RackUnit's non-UI code is modularized. What about if `rackunit` provides the meta checks and we open a separate issue for finer-grained modules?

To me, the `(Treeof FailurePred)` argument API is too complex. It presents multiple ways of doing the same thing, e.g. these are all nearly equivalent:

```racket
(test-begin
  (check-fail (list #rx"foo" pred) thnk))

(test-begin
  (check-fail #rx"foo" thnk)
  (check-fail pred thnk))

(test-begin
  (check-fail (list (list #rx"foo") (list (list pred))) thnk))
```

One of these could be preferred, but will they have subtle differences that a user has to remember? What about the extra location info provided by separate checks? And when using only a single predicate / regex / info, does it needs to be wrapped in a list to be considered a tree? Can you mix predicates / regexes / infos in the same tree? Is the order important? Is the tree traversed depth-first or breadth-first? Why is it a tree and not a list? These are all questions I'd have to ask myself as a user of this API, and I'd likely forget the answers to some of them frequently and have to consult the docs. Not to mention there's the increased implementation and contract complexity. I don't see the benefits of this approach outweighing the costs.

bennn:
I like @rmculpepper 's suggestions.

- `:+1:` for `rackunit/meta`, because it distinguishes "RackUnit for the meta-tester" from "RackUnit for the normal user".
  - it's fine with me to add `rackunit/base` etc. as long as `(require rackunit)` still gives the same imports. 
- `:+1:` for having one form
- `:+1:` for accepting a tree. The `plot` library has a similar (very convenient) API for renderers. I think it should check everything in the tree & report all errors. Order should be unspecified --- if you want to meta-meta-test then only supply non-list arguments

jackfirth:
Re `rackunit/meta`: I will move these exports to `rackunit/meta` and *not* export them from `rackunit`. This PR will not attempt to add a `rackunit/base` module or make any changes to the current exports of `rackunit`.

As for the tree API:

A tree API wouldn't quite be a single form, because it would still need separate forms for normal failures and errors. A third form that that supplied an empty tree by default would also be needed if we wanted to match the simplicity of `check-fail*`.

A tree of renderers in `plot` makes sense because there would be a lot of duplication otherwise, and there's no semantic changes that can occur as a result of the order of renderers in the tree. Changing the order of values in the tree supplied to a tree-based meta check form would change which error is shown, and I don't see any circumstance where the `message` argument to such a check would apply to all leaves of a tree.

A tree API also makes it harder to distinguish between different kinds of failures. Each leaf of the tree would have the same check location information. DrRacket's integration with the location information makes it easy to instantly see exactly which cases failed; I consider it bad practice for a rackunit API to encourage users to subvert that feature. Maybe if we had some way to point to different leaves of the tree in the location message?

bennn:
I still like the tree API

- I don't mind losing `check-fail*`
- RackUnit can test everything in the tree (so changing order only changes the orders of errors, not "which errors are shown")
- to distinguish kinds of errors, you can split one `check-fail*` into multiple (`check-exn` has the same "problem" --- a user can write one big predicate that tests lots of things & is hard to debug)

jackfirth:
I'd like to get this feature in and it seems I'm the only one doubting a tree API, so I shall defer to your judgment. Tree API it is!

With multiple errors, how should the checks report failures? Multiple failures for one check is something I've been working on in [racket-expect](https://github.com/jackfirth/racket-expect) and it can be complex enough that I'd rather avoid implementing it entirely here.

bennn:
I think just normal messages separated by whitespace would be OK. Example:

```
--------------------
FAILURE
name:       check-equal?
location:   foo.rkt:4:0
actual:     1
expected:   0

name:       check-equal?
location:   foo.rkt:5:0
actual:     1
expected:   2
--------------------
```

But since name & location are always the same, maybe each failed check in the tree could register a unique check-info key & all these keys could be printed on failure.

jackfirth:
Just printing like this would be fine with me:

```
--------------------
FAILURE
name:       check-fail?
location:   foo.rkt:4:0
actual:     1
expected:   0
actual:     1
expected:   2
--------------------
```

Anything more complicated I'd defer to a separate package designed with multiple failures in mind.

bennn:
Can simplify 1 more step and only print `actual` 1 time?
So `check-fail` prints one actual & multiple `expected`s

jackfirth:
This is now ready for another review.

Now, the only export of `rackunit/meta` is `check-fail`. It accepts a tree of predicates, regexps, and check info values. Here is a sample use and it's output:

```
(define (fail-foo)
  (with-check-info (['foo 1])
    (fail-check "some message from foo")))

(define (has-bar-info? exn)
  (member 'bar (map check-info-name (exn:test:check-stack exn))))

(check-fail (list has-bar-info?
                  #rx"message from bar"
                  (make-check-info 'bar 1))
            fail-foo)

--------------------
FAILURE
name:            check-fail
location:        unsaved-editor:13:0
actual:          (exn:test:check "some message from foo" #<continuation-mark-set> ...)
actual-message:  "some message from foo"
actual-info:
  foo:             1
expected:
  predicate:       #<procedure:has-bar-info?>
  message:         #rx"message from bar"
  info:            (check-info 'bar 1)
--------------------
```

The `check-fail/error` form is no longer needed because #75 means check failures evaluated in thunks given to `check-exn` will raise their exceptions normally and not log failures (hooray!)

See the `meta-test.rkt` module for a more substantial example of how one would test custom rackunit checks.

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -456,4 +457,51 @@ Raises an @racket[exn:test:check] with the contents of the @tech{check-info stac
...
+
+@defproc[(check-fail [assertion-tree
+                      (treeof (or/c (-> exn:test:check? any/c)
+                                    regexp?
+                                    check-info?))]
+                     [thunk (-> any)]
+                     [message string? ""])
+         void?]{
+ Checks that @racket[thunk] raises a check failure and that the failure
+ satisfies @racket[assertion-tree]. The tree is checked in the following manner:
```
bennn: I think the wording here is too low-level (sounds too much like an algorithm to me).

suggestion: change this line & the list to say

> satisfies every assertion in `assertion-tree`. A failure can satisfy an assertion in one of three ways, depending on the type of the assertion:
> - a failure `f` satisfies a predicate `p` if `(p f)` is a true value
> - a failure satisfies a regular expression `r` if `r` matches the failure's message,
> - a failure satisfies a `check-info` structure if the failure's check stack contains the expected value

... okay I don't like what I wrote very much either. Maybe it'd be better to have a `deftech` for the `(treeof ....)` contract and define "satisfies" in that?

Anyway, what's important to me is that the 3 cases for an "assertion" have 3 clear meanings. (I don't think the 4-point list & focus on "satisfying a tree" does that.)

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -456,4 +457,51 @@ Raises an @racket[exn:test:check] with the contents of the @tech{check-info stac
...
+
+ @(itemlist
```
bennn: `@itemlist[ .... ]`

(same for the `examples` below)

rackunit-lib/rackunit/meta.rkt
```diff
@@ -0,0 +1,115 @@
...
+                  pretty-info))
+
+
+(define-check (check-fail tree chk-thnk)
+  (contract-tree! 'check-fail tree)
+  (contract-thunk! 'check-fail chk-thnk)
+  (define failure (check-raise-value chk-thnk))
+  (unless (exn:test:check? failure)
+    (with-actual failure
+      (fail-check "Check raised error instead of failing")))
```
bennn: replace "failing" with "calling `fail-check`"?

Because I think:
- "failing" is a technical term, it really means to throw an `exn:test:check`
- "failing" isn't described in the docs (anyway, someone writing their first `define-check` will have trouble with the current error message)

rackunit-lib/rackunit/meta.rkt
```diff
@@ -0,0 +1,115 @@
...
+  (check-tree-assert tree failure))
+
+;; Shorthands for adding infos
+
+(define-simple-macro (with-actual act:expr body:expr ...)
+  (with-check-info* (error-info act) (λ () body ...)))
+
+(define (list/if . vs) (filter values vs))
```
bennn: Can this go in a common place?

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -456,4 +457,51 @@ Raises an @racket[exn:test:check] with the contents of the @tech{check-info stac
...
+
+Custom checks such as those created by @racket[define-check] can contain a fair
+amount of logic. Consequently, custom checks can be buggy and should be tested.
+RackUnit provides a few checks explicitly designed for testing the behavior of
+other checks; they allow verifying checks pass and fail when expected or that
+checks add certain information to the check information stack. These bindings
+are provided by @racketmodname[rackunit/meta], not @racketmodname[rackunit].
+
+@defproc[(check-fail [assertion-tree
+                      (treeof (or/c (-> exn:test:check? any/c)
```
bennn: I guess `any/c` should be `any`

rackunit-lib/rackunit/meta.rkt
```diff
@@ -0,0 +1,115 @@
...
+               (procedure-arity-includes? thnk 0))
+    (raise-argument-error name "(-> any)" thnk)))
+
+(define (contract-tree! name tree)
+  (for ([v (in-list (flatten tree))])
+    (unless (or (and (procedure? v)
+                     (procedure-arity-includes? v 1))
+                (regexp? v)
+                (check-info? v))
+      (define ctrct "(or/c (-> any/c boolean?) regexp? check-info?)")
```
bennn: `boolean?` => `any`

rackunit-test/tests/rackunit/meta-test.rkt
```diff
@@ -0,0 +1,45 @@
...
+                (λ () (check-fail #rx"foo" fail-not-foo)))
+    (check-fail (make-check-info 'expected (pretty-info some-info))
+                (λ () (check-fail some-info fail-check)))
+    (check-exn exn:fail:contract?
+               (λ () (check-fail 'nonsense fail-check)))
+    (check-exn exn:fail:contract?
+               (λ () (check-fail (list #rx"foo" 'partial-nonsense)
+                                 fail-check)))
+    (check-exn exn:fail:contract?
+               (λ () (check-fail accepts-no-args fail-check)))))
```
bennn: Add some tests with silly lists?

like `'(((#rx"hello") ()) #rx"world")` ... just non-flat variations on the tests you already have

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: The predicates have to return one value so I think `any/c` is appropriate here.

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: Done

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -456,4 +457,51 @@ Raises an @racket[exn:test:check] with the contents of the @tech{check-info stac
...
 message.}
 
+@section{Testing Custom Checks}
+@defmodule[rackunit/meta]
+
+Custom checks such as those created by @racket[define-check] can contain a fair
+amount of logic. Consequently, custom checks can be buggy and should be tested.
+RackUnit provides a few checks explicitly designed for testing the behavior of
+other checks; they allow verifying checks pass and fail when expected or that
+checks add certain information to the check information stack. These bindings
```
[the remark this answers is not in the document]
jackfirth: Done

rackunit-lib/rackunit/meta.rkt
jackfirth: I've written this in half a dozen different codebases so really I'd prefer to see it added to `racket/list` or some a `list-more` package. But I'll put it in `rackunit/private/util-list` for now (not `util`, because `rackunit/private/check` duplicates this and it can't depend on `util`).

rackunit-lib/rackunit/meta.rkt
jackfirth: Changed to `any/c` (see earlier comment about `any` vs `any/c`)

rackunit-lib/rackunit/meta.rkt
jackfirth: Done, very good idea.

rackunit-test/tests/rackunit/meta-test.rkt
jackfirth: Done. I think one test that uses all three kinds of assertions in a single silly list ought to be enough; more tests than that seem redundant.

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -1,10 +1,11 @@
 #lang scribble/doc
-@(require "base.rkt")
+@(require (except-in "base.rkt" examples)
+          scribble/example)
 
 @(require (for-label racket/match))
```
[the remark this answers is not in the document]
jackfirth: It's in `plot` somewhere, but I'm wary of adding a dependency on the plot package's docs to `rackunit-doc`. Maybe `treeof` should be moved out of the plot package?

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: I like what you wrote and definitely agree about having three list cases instead of four. I used slightly different wording though.

A `deftech` seems out of place here, especially since it's only used in this one check.

jackfirth:
Sorting and grouping infos in the failure message is now implemented.

rackunit-lib/rackunit/meta.rkt
```diff
@@ -0,0 +1,121 @@
...
+;; Shorthands for adding infos
+
+(define-simple-macro (with-actual act:expr body:expr ...)
+  (with-check-info* (error-info act) (λ () body ...)))
+
+(define (error-info raised)
+  (define (exn-info) (make-check-info 'exn (pretty-info raised)))
+  (define (msg-info) (make-check-info 'message (exn-message raised)))
+  (define (info-info)
+    (make-check-info 'info (nested-info (exn:test:check-stack raised))))
```
bennn: I wish the `actual` and `expected` info fields looked more similar:

```
actual:
  exn:        (exn:test:check "some message from foo" #<continuation-mark-set> ...)
  message:    "some message from foo"
  info:
    foo:        1
expected:
  predicate:  #<procedure:has-bar-info?>
  message:    #rx"message from bar"
  info:       (check-info 'bar 1)
```

Because, what if `foo` isn't a `check-info` struct, but something else that prints the same way?

I'd like to do `(nested-info (map pretty-info (exn:test:check-stack raised)))`, but that's a contract error.

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -490,4 +492,56 @@ Raises an @racket[exn:test:check] with the contents of the @tech{check-info stac
...
+   @racket[equal?] to @racket[i].}]
+
+ @examples[#:eval rackunit-eval
+           (check-fail '() (λ () (check-equal? 'foo 'bar)))
+           (check-fail number? (λ () (check-equal? 'foo 'bar)))
+           (check-fail (list string? (check-info 'info 10))
+                       (λ () (check-equal? 'foo 'foo)))]
+
+ Note that a single predicate, regular expression, or @tech{check-info} value is
+ considered a legal tree, as well as the empty list.
```
bennn: I don't think the docs really need this note.

But maybe, add a reminder note not to abuse to tree API? (A check should test "1 thing".)

rackunit-lib/rackunit/meta.rkt
```diff
@@ -0,0 +1,121 @@
...
+      (fail-check "Check raised error instead of calling fail-check")))
+  (check-tree-assert tree failure))
+
+;; Shorthands for adding infos
+
+(define-simple-macro (with-actual act:expr body:expr ...)
+  (with-check-info* (error-info act) (λ () body ...)))
+
+(define (error-info raised)
+  (define (exn-info) (make-check-info 'exn (pretty-info raised)))
```
bennn: I think these thunks are more time/space expensive than just doing `(define exn-info (make....`

jackfirth: The optimizer should be able to inline them, so I don't think that's worth worrying about.

rackunit-lib/rackunit/meta.rkt
jackfirth: Do you mean what if `foo` is a `check-info` struct who's value is not `1`, but something that prints the same such as `(string-info "1")`? Calling `exn:test:check-stack` will always return a list of `check-info` values modulo unreasonable shenanigans.

bennn: I didn't know that about `exn:test:check-stack`.

Okay though, my real complaint is that `foo:        1` and `(check-info 'bar 1)` look too different. I'd rather the first looked like `(check-info 'foo 1)`.

PR #61, opened 2017-07-15
Add replace-check-info and replace-check-info*

### LINKED ISSUE #58: How to update check-info location?
Here's a macro that uses `with-check-info*` to set a check's location relative to its argument.

```
#lang racket/base

  (syntax-parse stx
   [(_ . e*)
    #`(begin .
                (λ () (check-true #,e)))))]))

(check-true*
 #true
 #false)
```

Running this on HEAD prints an error with 2 `location:`s
```
--------------------
FAILURE
location:   test.rkt:14:1
name:       check-true
location:   test.rkt:10:22
params:     '(#f)
--------------------
```

Running on 6.5 prints only one:
```
--------------------
FAILURE
name:       check-true
expression: (check-true #f)
params:     (#f)
location:   (#<path:/Users/ben/code/racket/gtp/shallow/model/test.rkt> 14 1 387 6)

Check failure
--------------------
```

Should I change my macro?

jackfirth:
Closes #58. Chose the name `replace-check-info` instead of `update-check-info` to make it clearer that existing infos are being changed in-place.

The `replace-check-info` macro takes the info name as a plain identifier rather than an expression the way `with-check-info`does. The expression behavior of `with-check-info` always seemed like a bizarre misfeature to me, and I constantly make the mistake of writing `(with-check-info ([foo 'foo]) ...)`. However, I'm on the fence about making `replace-check-info` inconsistent with `with-check-info`.

### DIFF
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -369,7 +369,9 @@ misspelling errors:
    (define-check (fail/replaced-name)
      (replace-check-info [name "replaced!!!"]
        (fail-check)))
-   (fail/replaced-name))}
+   (fail/replaced-name))
+
+ @history[#:added "1.8"]}
 
 @defproc[(replace-check-info* [info check-info?] [thunk (-> any)]) any]{
  Like @racket[update-check-info], but as a normal procedure instead of a macro.
@@ -378,7 +380,9 @@ misspelling errors:
    #:eval rackunit-eval
    (define-check (fail/replaced-params)
      (replace-check-info* (make-params-info "replaced!!!") fail-check))
-   (fail/replaced-params))}
+   (fail/replaced-params))
+
+ @history[#:added "1.8"]}
 
 @section{Custom Checks}
 
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -353,7 +353,33 @@ misspelling errors:
              (make-check-info 'bar 2)))
      (with-check-info* custom-info fail-check))
    (fail/proc-info))}
-     
+
+@defform[(replace-check-info [name-id val-expr] body ...+)
+         #:contracts ([val-expr any/c])]{
+ Replaces the value of the current check info named @racket[name-id] with
+ @racket[val-expr] in the dynamic extent of the @racket[body] expressions. If no
+ check info named @racket[name-id] is present, appends a new check info to the
+ end of the stack. If multiple check infos named @racket[name-id] are present,
+ the first info has its value replaced and the other infos are removed. Unlike
+ @racket[with-check-info], @racket[name-id] is a plain identifier instead of an
+ expression producing a symbol.
+
+ @(examples
+   #:eval rackunit-eval
+   (define-check (fail/replaced-name)
+     (replace-check-info [name "replaced!!!"]
+       (fail-check)))
+   (fail/replaced-name))}
+
+@defproc[(replace-check-info* [info check-info?] [thunk (-> any)]) any]{
+ Like @racket[update-check-info], but as a normal procedure instead of a macro.
+
+ @(examples
+   #:eval rackunit-eval
+   (define-check (fail/replaced-params)
+     (replace-check-info* (make-params-info "replaced!!!") fail-check))
+   (fail/replaced-params))}
+
 @section{Custom Checks}
 
 Custom checks can be defined using @racket[define-check] and
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -1,5 +1,6 @@
 #lang scribble/doc
-@(require "base.rkt")
+@(require (except-in "base.rkt" examples)
+          scribble/example)
 
 @(require (for-label racket/match))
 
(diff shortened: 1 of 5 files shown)


rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -314,47 +315,74 @@ misspelling errors:
...
-@racket[thunk]}
-
-@interaction[#:eval rackunit-eval
-  (with-check-info*
-   (list (make-check-info 'time (current-seconds)))
-   (lambda () (check = 1 2)))
-]
+@defform[(with-check-info ([name-expr val-expr] ...) body ...+)
+         #:contracts ([name symbol?] [val-expr any/c])]{
+ Appends the given check infos to the check information stack for the dynamic
```
bennn: Can you make a `@deftech` for the "check information stack"

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -314,47 +315,74 @@ misspelling errors:
...
+ @(examples
+   #:eval rackunit-eval
+   (define-check (fail/proc-info)
+     (define custom-info
+       (list (make-check-info 'foo 1)
+             (make-check-info 'bar 2)))
+     (with-check-info* custom-info fail-check))
+   (fail/proc-info))}
+
+@defform[(replace-check-info [name-id val-expr] body ...+)
```
bennn: I'd rather call it `set-check-info`, to analogy with `hash-set`

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -314,47 +315,74 @@ misspelling errors:
...
+         #:contracts ([val-expr any/c])]{
+ Replaces the value of the current check info named @racket[name-id] with
+ @racket[val-expr] in the dynamic extent of the @racket[body] expressions. If no
+ check info named @racket[name-id] is present, appends a new check info to the
+ end of the stack. If multiple check infos named @racket[name-id] are present,
+ the first info has its value replaced and the other infos are removed. Unlike
+ @racket[with-check-info], @racket[name-id] is a plain identifier instead of an
```
bennn: I don't like this --- should stay consistent with `with-check-info`

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -314,47 +315,74 @@ misspelling errors:
...
+     (replace-check-info [name "replaced!!!"]
+       (fail-check)))
+   (fail/replaced-name))
 
-@verbatim{current-element: 8}
+ @history[#:added "1.8"]}
 
-is displayed along with the usual information on an check failure.
+@defproc[(replace-check-info* [info check-info?] [thunk (-> any)]) any]{
+ Like @racket[update-check-info], but as a normal procedure instead of a macro.
```
bennn: typo: `update-check-info`

rackunit-doc/rackunit/scribblings/check.scrbl
bennn: I also think this function should accept a list, like `with-check-info`

rackunit-lib/rackunit/private/check-info.rkt
```diff
@@ -59,12 +63,29 @@
...
-(define-syntax with-check-info
-  (syntax-rules ()
-    [(_ ((name val) ...) body ...)
-     (with-check-info*
-         (list (make-check-info name val) ...)
-       (lambda () body ...))]))
+(define-simple-macro (with-check-info ([name:expr val:expr] ...) body ...+)
+  (with-check-info* (list (make-check-info name val) ...) (λ () body ...)))
+
+(define (member-index v lst equal?)
```
bennn: I think you can remove this and use `index-of` instead

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: Re: list vs single info - what do you think the right thing to do is if the list of infos contains multiple infos with the same name? If only a single info is accepted, replacing multiple infos requires nesting expressions which makes it clear that the later infos replace the earlier ones. Should that be what happens if a list with those infos is given?

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: This seems a little confusing, because if changed to accept a list of infos instead of a single info someone could interpret this more like `set!` and think it removes _all_ existing check infos from the stack, setting the entire stack to the given list of infos.

bennn: This doesn't worry me ... (1) its just like `hash-set` (2) there's no `!` in the name (3) the documentation will make everything clear

rackunit-doc/rackunit/scribblings/check.scrbl
bennn: It should do what `with-check-info` used to do.

I'm not sure what that was, but I guess `(with-check-info (list A) (with-check-info (list B) ....` would behave the same as `(with-check-info (list A B) ....`

bennn:
The more I think about this, I'd wish:
- the `check-info` stack works like a hashtable (no duplicate keys) 
- RackUnit doesn't add `replace-check-info`
- and `racket-expect` uses uninterned symbols to get multiple "faults" on the stack

jackfirth:
I think you're right. Using uninterned symbols is a good idea, that should work fine for my purposes. Treating the check info stack like an assoc list seems simpler than this.

jackfirth:
Closing in favor of #64, which is way simpler. Thanks @bennn!

PR #66, opened 2017-07-31
doc: add deftech for check-info stack

bennn:
Define and use two deftechs:
- `check-info` (a key-value pair RackUnit uses to print error messages)
- `check-info stack` (a collection of `check-info`s that the user can extend)

I picked `check-info` rather than `check information` because that's how
 I'd rather talk about these things in real life.

I don't really like calling it a "stack", but I like saying "Pushes"
 in the doc for `with-check-info`.

### DIFF
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -335,7 +335,7 @@ is printed along with the usual information on an check failure.
 @defform[(with-check-info ((name val) ...) body ...)]{
 
 The @racket[with-check-info] macro pushes the given
-information in the @tech{check-info stack} for the duration
+information onto the @tech{check-info stack} for the duration
 of the execution of the body expressions. Each @racket[name] must be
 a quoted symbol and each @racket[val] must be a value.}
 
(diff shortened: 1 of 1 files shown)


rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -334,10 +334,10 @@ is printed along with the usual information on an check failure.
 
 @defform[(with-check-info ((name val) ...) body ...)]{
 
-The @racket[with-check-info] macro stores the given
-information in the check information stack for the duration
-of the execution of the body expressions.  @racket[Name] is
-a quoted symbol and @racket[val] is any value.}
+The @racket[with-check-info] macro pushes the given
+information in the @tech{check-info stack} for the duration
```
jackfirth: stores ... in -> pushes ... onto?

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -411,9 +413,9 @@ We can use these checks in the usual way:
           (define-binary-check (name actual expected) body ...)]]{
 
 The @racket[define-binary-check] macro constructs a check
-that tests a binary predicate.  It's benefit over
-@racket[define-simple-check] is in better reporting on check
-failure.  The first form of the macro accepts a binary
+that tests a binary predicate.  Compared to
+@racket[define-simple-check], does a better job reporting check
```
jackfirth: I think a word is missing after "Compared to `define-simple-check`"

PR #72, opened 2017-08-15
Simplify trim-current-directory implementation

jackfirth:


samth:
This has several differences from the current implementation; I don't know if they're important.

1. It works on paths as well as strings.
2. It has a problem in that sometimes `find-relative-path` can return a string (this seems like a mis-feature but we have to deal).
3. If `path` and `(current-directory)` are the same, the current implementation produces `""` but the new one produces `path`.

jackfirth:
I did this long enough ago that I no longer have any idea whatsoever whether those differences are important. Safest thing would probably be to close this.

PR #75, opened 2017-08-29
Move test-log logic for checks and tests inside their "around" parameters

### LINKED ISSUE #9: Wrapping a check with `define-check` reports two tests passed instead of one
If you do this:

``` racket
(define-check (check-string=? actual-str expected-str)
  (check string=? actual-str expected-str))

(check-string=? "foo" "foo")
```

And then run the test with `raco`, it reports "2 tests passed", rather than "1 test passed". This is problematic when making simple utility checks that don't do much more than provide a shorthand for a more configurable base check, as when building layers of simplifying checks the effects compound. For instance a `(check-foo? v)` wrapping a `(check-bar-equal? v foo)` wrapping a `(check-frobnicated-bars-equal? (frobnicate v) (frobnicate foo))` wrapping a `(check-equal? (id (frobnicate v)) (id (frobnicate foo)))` reports that 4 tests passed.

### LINKED ISSUE #14: Testing a check with a check

### LINKED ISSUE #58: How to update check-info location?

### LINKED ISSUE #70: Improve handling of `test-log!` and `current-check-around`
Today, checks log failures / successes using `test-log!` in a context outside `current-check-around`. This means the following expression:

```racket
(parameterize ([current-check-around (lambda (thnk) (thnk))])
  (check-equal? 1 2))
```

... *appears* to evaluate `(check-equal? 1 2)` without printing anything and with a check failure exception raised. But it has the hidden side effect of logging a failure with `test-log!`. This causes confusion resulting in issues like #14, and makes it difficult to implement a fix to #9.

Instead, I propose the following changes:

1. Call `test-log!` as part of `current-check-around`
2. Make `test-case` and `test-suite` log results once-per-case-or-suite instead of once-per-check, since the entire case or suite is conceptually a single test. If this isn't done, `test-case` and `test-suite` would need to reproduce the current behavior which could be tricky.
3. Make `define-check` use a different version of `current-check-around` in the body of `define-check` so nested checks raise exceptions, deferring failure printing and test logging to the outermost check's current check handler. This ought to solve #9.

jackfirth:
Closes #70 
Closes #9 
Closes #14 
Closes #58 

This PR moves the `test-log!` logic of checks and test cases into the `current-check-around` and `current-test-case-around` parameters, respectively. Additionally, checks now parameterize `current-check-around` inside the body of `define-check` so that *nested* checks are evaluated as normal functions. This allows checks to be *imperatively composed*; they can be chained together in series just like normal exception-throwing validation functions. This involves several subtle changes to the effects and evaluation of checks:

### Single nested check

```racket
(define-check (check-zero v)
  (check-equal? v 0))

(check-zero 0)
```

This now reports one test passing, and uses the default check info added by `check-zero`. Previously, it would report two tests passing.

### Multiple nested checks

```racket
(define-check (check-zero-twice v)
  (check-equal? v 0)
  (check-equal? v 0))

(check-zero-twice 0)
(check-zero-twice 1)
```

The passing use reports one test passing, and the failing use reports one test failing with the check failure thrown by the first `check-equal?` expression. Previously, the passing use reported two tests passing and the failing use reported two tests failing and one test passing, as each of the nested checks would raise and swallow a check failure leading the outer check to believe nothing went wrong.

### Multiple nested checks in test case

```racket
(test-case "passing test"
  (check-zero-twice 0))
(test-case "failing test"
  (check-zero-twice 1))
```

The passing test now reports one test passing, and the failing test reports one test failing. The passing test previously reported three tests passing, and the failing test previously reported one test passing *and* one test failing. It also only evaluated the first check inside `check-zero-twice`.

### Multiple checks in a test case

```racket
(test-case "passing test"
  (check-equal? 1 1)
  (check-equal? 2 2))
(test-case "failing test"
  (check-equal? 1 1)
  (check-equal? 2 3))
```

The passing test now reports one test passing, and the failing test now reports one test failing. Previously the passing test reported two tests passing and the failing test reported one test passing and two tests failing.

### Check in a test suite

For all test suites, calling `run-tests` on a suite now causes `test-log!` to report the same number of passing and failing tests as `run-tests` displays in its summary message. Previously, `test-log!` would report significantly more tests than `run-tests` claimed.

### Other changes

- Some parameters now use contracts instead of manually checking types
- Some of rackunit's own tests are now plain `test-case` expressions instead of `test-suite` definitions provided and combined in `all-rackunit-tests.rkt`
- The `rackunit/log` tests now have slightly better failure messages
- Calls to `test-log!` now need to be explicitly made by the user of `fold-test-results` or `foldts-test-results`. This also allows rackunit test suites to be run without interacting with `rackunit/log` at all (which the GUI runner probably wants to do anyway)

### Future work

This doesn't add any documentation of how rackunit checks behave when composed. It now mostly works "as expected", but without docs and tests guaranteeing this behavior it's not safe for users to rely on.

There are corresponding changes that need to be made in Typed Racket's [rackunit wrapper code](https://github.com/racket/typed-racket/tree/master/typed-racket-more/typed/rackunit) that duplicates rackunit's test case and test suite macros.

### DIFF
--- a/rackunit-lib/rackunit/private/check.rkt
+++ b/rackunit-lib/rackunit/private/check.rkt
@@ -81,7 +81,6 @@
 ;; given exception.  Useful for propogating internal
 ;; errors to the outside world.
 (define (refail-check exn)
-  (test-log! #f)
   (raise
    (make-exn:test:check (exn-message exn)
                         (exn-continuation-marks exn)
--- a/rackunit-lib/rackunit/private/check.rkt
+++ b/rackunit-lib/rackunit/private/check.rkt
@@ -51,14 +51,11 @@
 ;; plain-check-around.
 (define (check-around thunk)
   (define handler (current-check-handler))
-  (with-handlers ([(λ (_) #t) handler]) (thunk))
-  (void))
+  (with-handlers ([(λ (_) #t) handler]) (thunk)))
 
 ;; Evaluates a check just like a normal function, with no calls to test-log!
 ;; or the current check handler. Check failures are raised as plain exceptions.
-(define (plain-check-around chk-thunk)
-  (chk-thunk)
-  (void))
+(define (plain-check-around chk-thunk) (chk-thunk))
 
 ;; This is the default for current-check-around, and ensures a check logs
 ;; test results appropriately.
--- a/rackunit-lib/rackunit/private/check.rkt
+++ b/rackunit-lib/rackunit/private/check.rkt
@@ -14,12 +14,13 @@
 
 (provide
  (contract-out
-  [fail-check (->* () (string?) void?)]))
+  [fail-check (->* () (string?) void?)]
+  [current-check-handler (parameter/c (-> any/c any))]
+  [current-check-around (parameter/c (-> (-> void?) any))]
+  [plain-check-around (-> (-> void?) void?)]))
 
-(provide current-check-handler
-         check-around
-         current-check-around
-
+(provide check-around
+         
          define-check
          define-binary-check
          define-simple-check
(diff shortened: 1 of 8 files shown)


rackunit-lib/rackunit/private/check.rkt
```diff
@@ -41,32 +42,40 @@
...
-(define current-check-around
-  (make-parameter
-   (λ (thunk) (check-around thunk) (void))
-   (lambda (v)
-     (if (procedure? v)
-         v
-         (raise-type-error 'current-check-around "procedure" v)))))
+  (define handler (current-check-handler))
+  (with-handlers ([(λ (_) #t) handler]) (thunk))
+  (void))
```
bennn: Does Typed Racket not depend on the return type of `check-around`?
(Or are you going to change TR before merging this?)

rackunit-lib/rackunit/private/test-case.rkt
```diff
@@ -44,23 +46,21 @@
...
-                             (and (exn:fail? e)
-                                  (not (exn:test? e))))
-                           (λ (e)
-                             (test-log! #f)
-                             (raise e))])
-          (parameterize
-              ([current-check-handler raise]
-               [current-check-around  check-around])
+          (parameterize ([current-check-around plain-check-around])
             (void)
```
bennn: remove this?

jackfirth: Removing it causes some tests to fail because `(test-begin)` was never a syntax error, but `(parameterize ([param v]))` is.

rackunit-lib/rackunit/private/check.rkt
jackfirth: Checks created with `define-check` always returned void, so this shouldn't change the types. That's still the case according to the contract so these void expressions can be removed. I'll do that before merging.

rackunit-lib/rackunit/private/test-case.rkt
bennn: okay, maybe change to `(void expr ....)` ?

jackfirth: Turns out that doesn't work because the `expr` expressions actually aren't expressions, they could be definition. I've added a comment and renamed the pattern variable to hopefully make this less confusing.

jackfirth:
@bennn I'll merge this later today unless you've got any more comments. Thanks for the review!

PR #77, opened 2017-09-03
Add rackunit-typed package

### LINKED ISSUE #63: Migrate `typed/rackunit` to this repository
The `typed-racket-more` package in the [Typed Racket repository](https://github.com/racket/typed-racket) defines a typed interface to RackUnit. This interface depends on some internal implementation details of RackUnit, as it essentially reimplements the testing macros with type annotations. In order to better support future changes to RackUnit, that typed wrapper could be moved to this repository by doing the following:

- Create a `rackunit-typed` package which provides the typed wrapper modules currently in `typed-racket-more`.
- Remove the typed RackUnit code from `typed-racket-more` and have `typed-racket-more` depend on `rackunit-typed` while listing it as an `implies` package.

This has the added benefit of allowing packages to use RackUnit with Typed Racket without pulling in all the dependencies of `typed-racket-more`, which include the database library, the web server library, multiple GUI packages, and various other odds and ends.

jackfirth:
Adds a `rackunit-typed` package with the contents of the `typed/rackunit` collection in the `typed-racket-more` package. This allows clients to avoid depending on `typed-racket-more` when all they need is typed unit tests, as `typed-racket-more` has a lot of heavyweight dependencies. This also makes it easier to change rackunit's code without breaking TR. After this is merged, I will add a `rackunit-typed` package to the catalog and open a pull request in TR to remove the rackunit code and add `rackunit-typed` to its `'implies` deps.

Closes #63

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -20,11 +20,13 @@ install:
 - echo file://`pwd`/pkgs-catalog/ > catalog-config.txt
 - raco pkg config catalogs >> catalog-config.txt
 - raco pkg config --set catalogs `cat catalog-config.txt`
-- raco pkg install --deps search-auto $PKG-test
-- raco pkg install --deps search-auto compiler-lib
+- raco pkg install --auto $PKG-test
+- raco pkg install --auto $PKG-typed
+- raco pkg install --auto compiler-lib
 - ls $HOME/.racket/download-cache
 
 script:
 - raco test --drdr -p $PKG-test
+- raco test --drdr -p $PKG-typed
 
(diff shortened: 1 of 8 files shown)


bennn:
This looks good.

I think you should make all 3 changes at once, but I agree that merging this PR is safe.

AlexKnauth:
For racket version before this pull request, does it make sense to create a version exception for `rackunit-typed`, pointing to an empty package that just declares a dep on `typed-racket-more`?

PR #78, opened 2017-09-03
Add dynamic-info

### LINKED ISSUE #76: Computing check info values at the time a failure occurs
I'd like a way to add a value to the check info that is only computed when `fail-check` is called. This lets me add values to the check info stack that refer to mutable state or an external system property, such as `current-milliseconds`, while being sure that the check error message contains the value that was current when `fail-check` was called. A `dynamic-info` struct similar to the `string-info` and `nested-info` structs could be added, and it could accept a thunk that computes the info value to show.

```racket
(with-check-info (['time (dynamic-info current-milliseconds)])
  (test-case "some long test" ...)
  (test-case "some other long test" ...)
  (when (do-this-test?)
    (test-case "some possibly-not-executed test" ...)))
```

This could possibly be done by mucking about with `current-check-around` and `current-test-case-around`, but that's complex, error prone, and hard to test. Additionally, it is currently not possible to change `current-check-around` inside a test case or suite because `test-begin` parameterizes it inside the thunk passed to `current-test-case-around`.

jackfirth:
Closes #76 

Adds a `dynamic-info` struct that allows a check-info value to be computed at the time the stack is rendered into a string, rather than when the info is added to the stack. This makes it easy to add an info for a parameter such as `current-directory` to a group of tests without messing with `current-check-handler`.

### DIFF
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -315,7 +315,7 @@ the @racket[with-check-info*] function, and the
    #:eval rackunit-eval
    (with-check-info (['current-dir (dynamic-info current-directory)])
      (check-equal? 1 2)
-     (parameterize ([current-directory "/tmp"])
+     (parameterize ([current-directory (find-system-path 'temp-dir)])
        (check-equal? 1 2))))
 
  @history[#:added "1.9"]}
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -267,16 +267,19 @@ the @racket[with-check-info*] function, and the
 @racket[with-check-info] macro.
 
 @defstruct[check-info ([name symbol?] [value any]) #:transparent]{
- A @deftech[#:key "check-info"]{check-info structure} stores information associated with the context of the
- execution of a check. The @racket[value] is written in a check failure message
- using @racket[write] unless it is a @racket[string-info] value or a
- @racket[nested-info] value.
+ A @deftech[#:key "check-info"]{check-info structure} stores information
+ associated with the context of the execution of a check. The @racket[value]
+ is normally written in a check failure message using @racket[write], but the
+ @racketmodname[rackunit] library provides several special formatting wrappers
+ that can influence how the check info value is printed.
+
  @history[#:changed "1.6" "Changed from opaque to transparent"]}
 
 @defstruct*[string-info ([value string?]) #:transparent]{
  A special wrapper around a string for use as a @tech{check-info} value. When
  displayed in a check failure message, @racket[value] is displayed without
  quotes. Used to print messages instead of writing values.
+ 
  @(interaction
    #:eval rackunit-eval
    (define-check (string-info-check)
@@ -284,13 +287,15 @@ the @racket[with-check-info*] function, and the
                        ['message (string-info "hello world")])
        (fail-check)))
    (string-info-check))
+
  @history[#:added "1.2"]}
 
 @defstruct*[nested-info ([values (listof check-info?)]) #:transparent]{
- A special wrapper around a list of @tech{check-infos} for use as a @racket[check-info]
- value. A check info whose value is a nested info is displayed as an indented
- subsequence of infos. Nested infos can be placed inside nested infos, yielding
- greater indentation.
+ A special wrapper around a list of @tech{check-infos} for use as a
+ @racket[check-info] value. A check info whose value is a nested info is
+ displayed as an indented subsequence of infos. Nested infos can be placed
+ inside nested infos, yielding greater indentation.
+ 
  @(interaction
    #:eval rackunit-eval
    (define-check (nested-info-check)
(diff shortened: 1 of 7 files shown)


rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -267,39 +267,59 @@ the @racket[with-check-info*] function, and the
...
        (list (make-check-info 'foo "foo") (make-check-info 'bar "bar")))
      (with-check-info (['nested (nested-info infos)]) (fail-check)))
    (nested-info-check))
+ 
  @history[#:added "1.7"]}
 
+@defstruct*[dynamic-info ([proc (-> any/c)]) #:transparent]{
+ A special wrapper around a procedure that produces a value for a
+ @tech{check-info}. When a @racket[dynamic-info] is displayed in a check info
+ stack, @racket[proc] is called to determine what value to display.
```
bennn: question: what if the procedure returns a special formatting wrapper?

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -267,39 +267,59 @@ the @racket[with-check-info*] function, and the
...
 @defstruct[check-info ([name symbol?] [value any]) #:transparent]{
- A @deftech[#:key "check-info"]{check-info structure} stores information associated with the context of the
- execution of a check. The @racket[value] is written in a check failure message
- using @racket[write] unless it is a @racket[string-info] value or a
- @racket[nested-info] value.
+ A @deftech[#:key "check-info"]{check-info structure} stores information
+ associated with the context of the execution of a check. The @racket[value]
+ is normally written in a check failure message using @racket[write], but the
+ @racketmodname[rackunit] library provides several special formatting wrappers
+ that can influence how the check info value is printed.
```
bennn: I'm thinking these "special formatting wrappers" need a `deftech` and dedicated subsection.
Maybe also a `gen:check-info-value` generic interface.

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -267,39 +267,59 @@ the @racket[with-check-info*] function, and the
...
+@defstruct*[dynamic-info ([proc (-> any/c)]) #:transparent]{
+ A special wrapper around a procedure that produces a value for a
+ @tech{check-info}. When a @racket[dynamic-info] is displayed in a check info
+ stack, @racket[proc] is called to determine what value to display.
+
+ @(interaction
+   #:eval rackunit-eval
+   (with-check-info (['current-dir (dynamic-info current-directory)])
+     (check-equal? 1 2)
+     (parameterize ([current-directory "/tmp"])
```
bennn: suggestion: replace `"/tmp"` with `(find-system-path 'temp-dir)`

rackunit-test/tests/rackunit/all-rackunit-tests.rkt
```diff
@@ -2,7 +2,6 @@
 
 (require rackunit
          "check-test.rkt"
-         "format-test.rkt"
```
bennn: wait, why?

I think this PR shouldn't change the `test-suite` style. Just add new tests for dynamic-info.

rackunit-doc/rackunit/scribblings/check.scrbl
bennn: ^ for an answer, can you add an example where the `dynamic-info` proc returns a string-info or something?

rackunit-test/tests/rackunit/all-rackunit-tests.rkt
jackfirth: Grouping all the tests together into one suite makes `raco test --drdr` give less useful information, since it breaks test failures down by module and `run-tests` reports all tests in the suite as coming from the module where `run-tests` was called. So I don't think it's a good idea to use test suites unless there's a specific feature of test suites that's needed.

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: Good idea, done.

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: The special formatting wrapper is printed as it would be if it had not been returned from a `dynamic-info`. Added an example with `nested-info`.

rackunit-doc/rackunit/scribblings/check.scrbl
jackfirth: The `deftech` is a good idea. A generic interface I'm not so sure about, since these values don't know how to format themselves out of context. For example, a `nested-info` struct can't turn itself into a string because it doesn't know how far to indent each line.

bennn: oh, I was thinking "save for another PR" when I wrote that comment. Sorry I didn't actually write that.

rackunit-test/tests/rackunit/all-rackunit-tests.rkt
bennn: I'd do that in a separate PR (remove all test suites). But LGTM

jackfirth: I'll open an issue to remove the suites and do that work separately in the future. In this PR I'll keep it included since separating it out doesn't seem worth it.

PR #87, opened 2018-01-20
define-check: only add location if not already on stack

bennn:
Change `define-check` to only add a location for the check if the `current-check-info` stack does not have a location on it.

(I think this is what `check-true` and others used to do. I'm not sure if they did the same for other check-info items too.)

#### With this PR,
this program fails with a good error message:
```
#lang racket/base

  (syntax-parse stx
   [(_ . e*)
    #`(begin .
                (λ () (check-true #,e)))))]))

(check-true*
 #true
 #false)
```

#### Without this PR,
the above program needs to be rewritten something like this to get a good error message:
```
#lang racket/base

(define (call/updated-location loc thnk)
  (define old-around (current-check-around))
  (define (new-around check-thnk)
    (with-check-info* (list (make-check-location loc))
      (λ () (old-around check-thnk))))
  (parameterize ([current-check-around new-around])
    (thnk)))

  (syntax-parse stx
    [(_ . e*)
    #`(begin .
            #`(call/updated-location '#,(build-source-location-list e) (λ () (check-true #,e)))))]))

(check-true*
 #true
 #false)
```

- - -

In case we'd rather close this PR without merging, I looked in the main distribution for uses of `with-check-info` that updated the location. The only ones I found were in the `rackunit-abbrevs` and `typed-racket-test` collections. I can change those to do the `call/updated-location` dance.

### DIFF
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -12,16 +12,18 @@
 Checks are the basic building block of RackUnit.  A check
 checks some condition and always
 evaluates to @racket[(void)].  If the condition doesn't hold, the
-check will report the failure (see @racket[current-check-handler]
-for customizing how failures are handled).
+check will report the failure using the current @tech{check-info stack}
+(see @racket[current-check-handler] for customizing how failures are handled).
 
 Although checks are implemented as macros, which is
-necessary to grab source location, they are conceptually
+necessary to grab source locations (see @secref{rackunit:custom-checks}), they are conceptually
 functions (with the exception of @racket[check-match] below).
 This means, for instance, checks always evaluate
-their arguments.  You can use checks as first class
-functions, though you will lose precision in the reported
-source locations if you do so.
+their arguments.  You can use a check as a first class
+function, though this will affect the source location that the check grabs.
+
+
+@section[#:tag "rackunit:basic-checks"]{Basic Checks}
 
 The following are the basic checks RackUnit provides.  You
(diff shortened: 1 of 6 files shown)


bennn:
The PR & issues about `check-exn` ignoring its message argument are/were related to this (#65).

I'm thinking, as a general rule, the `checks` in RackUnit should not add things to the check info stack if there's matching info already on the stack. (Like default arguments to a function.)

stchang:
Many of my tests failed to report the right source location, due to a change that occurred between 6.10 and 6.11.

This pr fixes my issue.

jackfirth:
I think this is the right thing to do and think you've got the right rule here:

> I'm thinking, as a general rule, the checks in RackUnit should not add things to the check info stack if there's matching info already on the stack. (Like default arguments to a function.)

I likely won't have time to review this (or do anything else with RackUnit) this week or weekend but I'm planning on getting to it by the end of the month. If before then you'd like to make a PR that makes all "automatic" check infos added by `define-check` behave like this, I have nothing but encouragement.

bennn:
I changed the docs (and code) to say that `define-check` adds `'location` and `'expression` if they aren't already there.

This is different from the rule I proposed above, but I think it makes sense based on todays reading of the documentation. I'll re-read next weekend.

bennn:
Updated,
- "automatic" check-info values are now only added if not already on the stack
- the docs for "Custom Checks" try to explain this
- added a new function `with-default-check-info*` --- like `with-check-info*` but does not override existing infos

PR #90, opened 2018-02-10
doc: fix typos, change contract for 'message' arguments

bennn:
If the 'message' is `#f` there is no message, if 'message' is `""` then it gets printed as the message.n


bennn:
594ef4d41a8de6672e3c63810cd5dab80bf2deff

PR #93, opened 2018-03-11
add empty rackunit-typed package for earlier racket versions

AlexKnauth:
When earlier racket installations (6.10 and below) look for the `rackunit-typed` package, they currently find the main non-empty rackunit-typed package in the catalog. This creates a conflict with `typed-racket-more`, with them both trying to provide different versions of `typed/rackunit`. This pull request adds a dummy rackunit-typed package that earlier racket versions can point to.

The [catalog entry for `rackunit-typed`](https://pkgd.racket-lang.org/pkgn/package/rackunit-typed) should also have version exceptions for versions 6.10 and below pointing to the package source:
```
git://github.com/racket/rackunit?path=rackunit-typed#for-v-up-to-6.10
```

### DIFF
--- a/rackunit-typed/info.rkt
+++ b/rackunit-typed/info.rkt
@@ -1,0 +1,10 @@
+#lang info
+
+(define collection 'multi)
+
+(define deps
+  '("base"))
+
+(define implies
+  '("typed-racket-more"))
+


samth:
Looks good to me.

AlexKnauth:
Can you add the version exceptions?

AlexKnauth:
@jackfirth You were the one to add the `rackunit-typed` package. Are you the only one who can add the version exceptions?

AlexKnauth:
Who can add version exceptions to the [`rackunit-typed` catalog entry](https://pkgs.racket-lang.org/package/rackunit-typed)?

The catalog says the only author is @jackfirth, but if he doesn't have time or can't for some other reason, who else can?

jackfirth:
Thank you for your patience @AlexKnauth. Life's gotten away from me lately.

1. I've added `samth@racket-lang.org` as an author of the `rackunit-typed` package. Is that alright with you @samth, and is that the right email address?
2. Not sure what's up with Travis CI, but I don't think there's anything wrong with the PR so I'll merge.
3. I'll add the version exceptions up to 6.0 after merging. Should I add fewer? Or add more by including the 5.9.x versions?

AlexKnauth:
I don't see those in [_Package Metadata_](https://docs.racket-lang.org/pkg/metadata.html) in the docs. What are the values supposed to be?

AlexKnauth:
There's a typo in the version exception for 6.9 in the [catalog entry](https://pkgd.racket-lang.org/pkgn/package/rackunit-typed). It refers to the (non-existent) `rackunit-type` repo where it  say `rackunit-typed`.

samth:
This is now fixed.

AlexKnauth:
Thanks.

Issue #99, opened 2018-08-08
`check-exn` gives internal error message when give the wrong number of arugments


florence:
```
#lang racket

(require rackunit)
(check-exn (lambda () 'hi))
```
Errors with:

```
rackunit-lib/rackunit/private/check.rkt:112:13: application: no case matching 1 non-keyword argument
  procedure: check-impl
  arguments...:
```

Which 1. doesn't talk about check-exn, 2. points into the internals of rackunit.

bennn:
https://github.com/racket/rackunit/commit/eceb061e974773a44bb0412df1eee06852e7d683

PR #100, opened 2018-08-11
report arity errors in terms of the public API

bennn:
Fix for #99 

Now this program:

```
#lang racket

(require rackunit)
(check-exn (lambda () 'hi))
```

Errors with:

```
check-exn: arity mismatch;
 the expected number of arguments does not match the given number
  given: 1
  arguments...:
   #<procedure:...rackunit/foo.rkt:4:11>
  context...:
   "/..../rackunit/foo.rkt": [running body]
   for-loop
   run-module-instance!125
   perform-require!78
```
- - -

This still needs a regression test. (EDIT: added a test)

But also, I'm not happy adding an extra `lambda` to get the keyword arguments. Suggestions welcome.

### DIFF
--- a/rackunit-lib/rackunit/private/check.rkt
+++ b/rackunit-lib/rackunit/private/check.rkt
@@ -90,36 +90,34 @@
 
 (define (list/if . vs) (filter values vs))
 
-(define-simple-macro
-  (define-check-func (name:id formal:id ...) #:public-name pub:id body:expr ...)
-  (define (name formal ... [message #f]
-                #:location [location (list 'unknown #f #f #f #f)]
-                #:expression [expression 'unknown])
-    (define infos
-      (list/if (make-check-name 'pub)
-               (make-check-location location)
-               (make-check-expression expression)
-               (make-check-params (list formal ...))
-               (and message (make-check-message message))))
-    (with-default-check-info* infos
-      (λ () ((current-check-around) (λ () body ... (void)))))))
+(define-simple-macro (make-check-func (name:id formal:id ...) #:public-name pub:id body:expr ...)
+  (λ (#:location [location (list 'unknown #f #f #f #f)]
+      #:expression [expression 'unknown])
+    (procedure-rename
+      (λ (formal ... [message #f])
+          (define infos
+            (list/if (make-check-name 'pub)
+                     (make-check-location location)
+                     (make-check-expression expression)
+                     (make-check-params (list formal ...))
+                     (and message (make-check-message message))))
+          (with-default-check-info* infos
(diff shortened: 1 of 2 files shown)


AlexKnauth:
Does this fix #103 as well?

rackunit-lib/rackunit/private/check.rkt
```diff
@@ -91,35 +91,36 @@
...
-               (make-check-expression expression)
-               (make-check-params (list formal ...))
-               (and message (make-check-message message))))
-    (with-default-check-info* infos
-      (λ () ((current-check-around) (λ () body ... (void)))))))
+  (make-check-func (name:id formal:id ...) #:public-name pub:id body:expr ...)
+  (procedure-rename
+    (λ (formal ... [message #f])
+      (λ (#:location [location (list 'unknown #f #f #f #f)]
+          #:expression [expression 'unknown])
```
AlexKnauth: Would it make sense to flip these two lambdas, making it more like:
```racket
  (λ (#:location [location (list 'unknown #f #f #f #f)]
      #:expression [expression 'unknown])
    (procedure-rename
      (λ (formal ... [message #f])
        ..stuff..)
      'pub))
```
?

It would simplify the `id` case of the `name` macro below.

bennn: I don't see how that would be simpler by itself. Wouldn't we just go from this:

```
((apply check-impl args) #:loc .... #:expr ....)
```

to this:

```
(apply (check-impl #:loc .... #:expr ....) args)
```

... maybe its best to remove `#:location` and `#:expression` and have `define-check` expand to code that updates `current-check-info` manually

bennn:
yes this fixes #103

rackunit-lib/rackunit/private/check.rkt
AlexKnauth: I mean the `id` case could be simplified from this:
```racket
          #'(λ args
               ((apply check-impl args)
                #:location (syntax->location #'loc)
                #:expression 'chk))
```
To this:
```racket
          #'(check-impl
              #:location (syntax->location #'loc)
              #:expression 'chk)
```

AlexKnauth: It also has the benefit of `(procedure-arity check-equal?)` returning `(list 2 3)` instead of `(arity-at-least 0)`.

bennn:
I'm planning to merge this tomorrow night

AlexKnauth:
To me, this looks good now.

bennn:
wow, great timing!

PR #101, opened 2018-08-16
rackunit/docs-complete: accept any `module-path?`

LiberalArtist:
This pull request updates the documentation to correspond to https://github.com/racket/racket/pull/2230


rfindler:
Thanks., but this needs an `@history` annotation.

bennn:
does it need a version bump too?

bennn:
#102

rfindler:
In principle, yes, but I don't think it really needs us to bother with a version bump, unless we know something is specifically depending on it (and wants to document that).

PR #104, opened 2018-08-20
test: raise reliable arity error

bennn:
change `(= 1)` to an expression that throws an arity error on 7.0.0.13 and earlier

### DIFF
--- a/rackunit-test/tests/rackunit/check-test.rkt
+++ b/rackunit-test/tests/rackunit/check-test.rkt
@@ -335,7 +335,11 @@
             (('name "name") ('info "info"))
           #t)
         (fail-check))))
-   
+
+  (define (raise-sample-arity-error)
+    ;; expected 1 argument, got 0
+    (raise-arity-error 'check-test 1))
+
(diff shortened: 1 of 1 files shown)


rackunit-test/tests/rackunit/check-test.rkt
```diff
@@ -335,7 +335,10 @@
             (('name "name") ('info "info"))
           #t)
         (fail-check))))
-   
+
+  (define (raise-sample-arity-error)
+    (raise-arity-error 'check-test 1))
```
samth: Why not call a function defined here, so that this never breaks?

bennn: that works too, but I thought this was clearer and should also never break

I guess it would be even clearer to define a new exception type ... `exn:fail:rackunit-test`. Do you think I should do that?

samth: Ah, I misunderstood -- I thought you were calling `raise-arity-error` with the wrong number of arguments to generate an arity error. All looks good.

PR #108, opened 2019-04-08
Change any to any/c where appropriate (fixes #106)

### LINKED ISSUE #106: any vs any/c?
Why does the documentation state that most test functions consume `any` rather than `any/c`[1]? In fact, the documentation of `any` seems to suggest that using `any` in this way will cause a syntax error.

[1]: See [`check-eq?`](https://docs.racket-lang.org/rackunit/api.html?q=test-suite#%28def._%28%28lib._rackunit%2Fmain..rkt%29._check-eq~3f%29%29) for instance.

sorawee:
Using `any` is a syntax error in non-return position.
This PR corrects this mistake in the documentation.
Note that the change is conservative: I only fix the cases that
are definitely incorrect, but there could still be incorrect
ones left.

### DIFF
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -130,7 +130,7 @@ For example, the following checks all fail:
 ]
 }
 
-@defproc[(check-exn (exn-predicate (or/c (-> any/c any/c) regexp?))
+@defproc[(check-exn (exn-predicate (or/c (-> any/c any) regexp?))
                     (thunk (-> any)) (message (or/c string? #f) #f))
          void?]{
 
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -130,7 +130,7 @@ For example, the following checks all fail:
 ]
 }
 
-@defproc[(check-exn (exn-predicate (or/c (-> any/c any) regexp?))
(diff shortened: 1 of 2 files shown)


sorawee:
Which `any` are you talking about? There are a few `any` that I didn't touch because they are in the return position, so they are fine.

bennn:
`test-case?`, `test-exn`, and `test-suite?` in the file `compound-testing.scrbl`

rackunit-doc/rackunit/scribblings/check.scrbl
```diff
@@ -130,7 +130,7 @@ For example, the following checks all fail:
 ]
 }
 
-@defproc[(check-exn (exn-predicate (or/c (-> any any/c) regexp?))
+@defproc[(check-exn (exn-predicate (or/c (-> any/c any) regexp?))
```
bennn: RackUnit currently expects the predicate to return one value:
```
> (check-exn (lambda (x) (values)) (lambda () (error 'die)))
--------------------
; ERROR [,bt for context]

result arity mismatch;
 expected number of values not received
  expected: 1
  received: 0
  values...:
--------------------
```

sorawee: Aha! I tested with 

```
(check-exn (lambda (x) (values)) (lambda () (values)))
```

so I thought it's OK. Will fix in a moment. Thanks :)

sorawee:
- `test-case?` and `test-suite?` are fixed in my second commit.
- `check-exn` and `test-exn` are fixed in my most recent commit.

PR #118, opened 2020-01-26
Use the installer from Utah in .travis.yml and run the cs version too

gus-massa:
The installer from Northwestern are failing, so use the installer from Utah instead. I added the `--fail` option to `curl` to get a better error message in case of a problem.

Run the `cs` version too. I think it's stable enough to avoid adding it to the `allow_failure` section.

I'm using the minimal installers. The full installers are a little faster, but I think in that case I should use `raco pkg update` or something like that.

I'm not sure how related is this to https://github.com/racket/rackunit/pull/110

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -8,11 +8,18 @@ cache:
 
 env:
 - PATH=~/racket/bin:$PATH
+- PATH=~/racket/bin:$PATH  VARIANT=cs
+
 
 before_install:
 - export PKG=`echo $TRAVIS_REPO_SLUG | cut -d '/' -f 2`
 - echo $PKG
-- curl -L -o installer.sh http://plt.eecs.northwestern.edu/snapshots/current/installers/min-racket-current-x86_64-linux-precise.sh
+- |
+    if [ "$VARIANT" = "cs" ]; then
+      curl --location --fail --output installer.sh http://www.cs.utah.edu/plt/snapshots/current/installers/min-racket-current-x86_64-linux-cs-xenial.sh;
+    else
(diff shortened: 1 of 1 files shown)


.travis.yml
```diff
@@ -8,22 +8,31 @@ cache:
...
 env:
 - PATH=~/racket/bin:$PATH
+- PATH=~/racket/bin:$PATH  VARIANT=cs
+
 
 before_install:
 - export PKG=`echo $TRAVIS_REPO_SLUG | cut -d '/' -f 2`
 - echo $PKG
-- curl -L -o installer.sh http://plt.eecs.northwestern.edu/snapshots/current/installers/min-racket-current-x86_64-linux-precise.sh
+# - curl -Lf -o installer.sh http://plt.eecs.northwestern.edu/snapshots/current/installers/min-racket-current-x86_64-linux-precise.sh
```
jackfirth: You can delete this instead of commenting it out, since it will be in the git history.

.travis.yml
```diff
@@ -8,22 +8,31 @@ cache:
...
+- |
+    if [ "$VARIANT" = "cs" ]; then
+      curl -Lf -o installer.sh http://www.cs.utah.edu/plt/snapshots/current/installers/min-racket-current-x86_64-linux-cs-xenial.sh;
```
jackfirth: Do the `L`, `f`, and `o` switches have long-form names? I prefer long form flags for scripts and config files, since they're read more than once. Same for the `-i` and `-v` flags down below.

gus-massa:
Done.

gus-massa:
Any other suggestion, or is it fine to merge this?

bennn:
merge!

PR #123, opened 2020-09-27
Fix incorrect failure report

sorawee:
Fixes: racket/racket#3410
CC: @AlexKnauth


jackfirth:
Fix looks good to me.

samth:
This change broke the following test: http://drdr.racket-lang.org/55531/cs/racket/share/pkgs/rackunit-test/tests/rackunit/standalone.rkt

sorawee:
Oh, interesting. Apparently `run-tests.rkt` didn't invoke `standalone.rkt`, so I didn't find the issue. Investigating now.

PR #124, opened 2020-10-08
Update test to reflect changes in #123 and #109

sorawee:
Sine we "treat exceptions thrown by arguments to check as test
failures", and test failures contain "params" clause, it makes sense
for the reported message to contain the clause as well.
The clause is also useful in itself.

Use procedure-rename to avoid names
like `#<procedure:.../path/to/file.rkt:x:y>` which is brittle and
especially too sensitive to changes.


sorawee:
CC: @AlexKnauth

samth (approved):

samth:
Is there a reason not to merge this now?

sorawee:
Nope, feel free to merge it whenever you feel appropriate.

PR #125, opened 2020-10-20
update current-check-around doc to match contract

stchang:
At some point, a contract was added to `current-check-around` (https://github.com/racket/rackunit/commit/1fbc98eeb7e3a6af06d7dbd32a6d6894e234297c#diff-be44254d70efbdd4192059f2a369dd9223835927001e849ed2e9628c5f63901e) but the doc was not updated.

Not 100% sure about `test-suite-check-around` but it seems to get set to `current-check-around`


stchang:
Actually, now I'm not sure the original change was correct.

PR #127, opened 2020-11-05
improve source locations in stacktraces

rfindler:
When there is an exception while running `check-equal?`, eg with this program:

```
#lang racket
(require rackunit)

(define (f x) (car x))
(set! f f)

(check-equal? 1 (f #f))
```

then this PR improves the error reporting so we see the source location of the `check-equal?` in the stack. Before this pull request, we'd see something like this:

```
car: contract violation
  expected: pair?
  given: #f
  context...:
   /Users/robby/git/exp/plt/racket/share/pkgs/rackunit-lib/rackunit/private/check.rkt:121:16
   body of "/Users/robby/tmp.rkt"
```

and after it, the line that points to `rackunit/private/check.rkt:121:16` points to the place where the `check-equal?` itself appears, `/Users/robby/tmp.rkt:7:0` in this case.

### DIFF
--- a/rackunit-lib/rackunit/private/check.rkt
+++ b/rackunit-lib/rackunit/private/check.rkt
@@ -113,15 +113,16 @@
       (with-syntax ([loc (datum->syntax #f 'loc stx)])
         (syntax-parse stx
           [(chk . args)
-           #'(let ([location (syntax->location #'loc)])
+           #`(let ([location (syntax->location #'loc)])
                (with-default-check-info*
                 (list (make-check-name 'name)
                       (make-check-location location)
                       (make-check-expression '(chk . args)))
-                (λ ()
-                  ((check-impl #:location location
-                               #:expression '(chk . args))
-                   . args))))]
+                #,(syntax/loc #'loc
+                    (λ ()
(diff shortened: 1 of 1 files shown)


rfindler:
Here's a test case for this behavior. It returns `#t` with the PR in place and `#f` without.

```
#lang racket
(require rackunit)

(define (f x) (car x))
(set! f f)

(define exn
  (with-handlers ([exn:fail? values])
    (check-equal? 1 (f #f))))

(for/or ([frame (in-list
                 (continuation-mark-set->context
                  (exn-continuation-marks exn)))])
  (match frame
    [(cons _ (srcloc pth _ _ _ _))
     (equal? pth (syntax-source #'here))]
    [_ #f]))
```

jackfirth:
Robby, this is fantastic.

rfindler:
Thanks, @jackfirth ! 

Do you know the right place to put the test case?

jackfirth:
@rfindler I think the [pr directory](https://github.com/racket/rackunit/tree/master/rackunit-test/tests/rackunit/pr) would be a good spot, with #127 as the PR number.

PR #133, opened 2021-01-31
typed: Remove extra call to `current-test-case-around`

### LINKED ISSUE #132: Incorrect test counting in typed/rackunit (test count doubled)
Working with `typed/rackunit` on 7.9, CS variant, macOS. Here's a rela

````
#lang typed/racket/base

(module+ test
  (require typed/rackunit))

(module+ test
  (test-case "one" (check-not-false 1)))
````

and then do `raco test`, you'll get

````
2 tests passed
````

If you add a test, so that we have two `test-case`s, like so:

````
#lang typed/racket/base

(module+ test
  (require typed/rackunit))

(module+ test
  (test-case "one" (check-not-false 1))
  (test-case "two" (check-not-false 2)))
````

you'll now get

````
4 tests passed
````

Looks like the count is always twice the number of checks. But there's more: if you make a failing test, like this:

````
#lang typed/racket/base

(module+ test
  (require typed/rackunit))

(module+ test
  (test-case "one" (check-not-false 1))
  (test-case "two" (check-not-false #f))) ; fails
````

Then you'll get:

````
--------------------
two
FAILURE
name:       check-not-false
location:
  /Applications/Racket v7.9/share/pkgs/rackunit-typed/rackunit/main.rkt:41:2
params:     '(#f)
--------------------
2/4 test failures
````

It looks like we're double counting the number of tests and double counting the number of failed tests.

The described issue doesn't arise in untyped `rackunit`.

jessealama:
It's not necessary to wrap a test case in `current-test-case-around`; that will be done later. Doing so increments the running test count (and, possibly, the running test failure count) twice.

The proposed change may seem implausible on its face. I also feel somewhat less than confident that this is the right thing to do. But I can confirm that this does the job.

closes #132

### DIFF
--- a/rackunit-lib/rackunit/private/check.rkt
+++ b/rackunit-lib/rackunit/private/check.rkt
@@ -21,7 +21,7 @@
   [plain-check-around (-> (-> void?) void?)]))
 
 (provide check-around
-         
+
          define-check
          define-binary-check
          define-simple-check
@@ -104,7 +104,7 @@
           (with-default-check-info* infos
             (λ () ((current-check-around) (λ () body ... (void))))))
       'pub)))
-                                      
+
 
 (define-simple-macro (define-check (name:id formal:id ...) body:expr ...)
   (begin
--- a/rackunit-typed/rackunit/main.rkt
+++ b/rackunit-typed/rackunit/main.rkt
@@ -97,17 +97,15 @@
   (syntax-case stx ()
     [(_ expr ...)
      (syntax/loc stx
-       ((current-test-case-around)
-        (lambda ()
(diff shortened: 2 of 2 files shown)


jackfirth:
I think the problem isn't that this macro expands to a use of `current-test-case-around`, I think it's that it tries to expand to `current-test-case-around` and also do the test result logging at the same time. Instead of removing the `current-test-case-around` usage, I think we should update this code to use roughly the same implementation strategy that the [untyped](https://github.com/racket/rackunit/blob/master/rackunit-lib/rackunit/private/test-case.rkt) macros use.

PR #140, opened 2021-05-04
Support images in check failure output

### LINKED ISSUE #130: Is there a way to have rackunit render picts inside the "check-info stack"?
Screen Shot 2020-12-26 at 23 13 00

## Problem
The "check-info stack" facility is unable to display visual data.

## Motivation

While working on a graph algorithm, I found myself wanting to visualize the graphs I was computing. `rackunit` makes it convenient to associate extra information with a test case failure through the "check-info stack". I've always used that facility to associate textual data with my test cases, but now I found myself wanting to add visual data to that.

## Workaround

Since I can only display text in test failures, I simply included the dot representation of the graph, like so:

Screen Shot 2020-12-27 at 14 31 33

aymanosman:
Fixes #130

This PR makes `rackunit` print values directly, as opposed to turning them into strings first.

### DIFF
--- a/rackunit-lib/rackunit/private/check-info.rkt
+++ b/rackunit-lib/rackunit/private/check-info.rkt
@@ -19,7 +19,7 @@
   [struct nested-info ([values (listof check-info?)])]
   [struct verbose-info ([value any/c])]
   [struct dynamic-info ([proc (-> any/c)])]
-  [info-value->string (-> any/c string?)]
+  [print-info-value (-> any/c any)]
   [current-check-info (parameter/c (listof check-info?))]
   [check-info-contains-key? (check-info-> symbol? boolean?)]
   [check-info-ref (check-info-> symbol? (or/c check-info? #f))]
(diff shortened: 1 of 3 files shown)


rmculpepper:
When I try running the following program in DrRacket, I still get the picts printed as `#<pict>` in the output. Is it working for you?

    #lang racket/base
    (require pict rackunit)
    (check-equal? (circle 10) (circle 20))

aymanosman:
You are right, thanks. I broke the implementation whilst fixing a merge conflict. Should be fixed now.

samth:
I think this merge broke the previous merge -- see the tests now failing at: http://drdr.racket-lang.org/57762/cs/racket/share/pkgs/rackunit-test/tests/rackunit/standalone.rkt and http://drdr.racket-lang.org/57762/cs/racket/share/pkgs/rackunit-test/tests/rackunit/run-tests.rkt

samth:
This change broke the `rackunit-pretty-runner` package, which was using the private `info-value->string` function. Would it make sense to restore that function using the new code written in this PR?

aymanosman:
@samth Can't seem to find that package anywhere.

samth:
@aymanosman it's here: https://pkgs.racket-lang.org/package/rackunit-fancy-runner

samth:
I see that you couldn't find it because I wrote the wrong name, sorry about that!

PR #142, opened 2021-05-31
Fix some failures discovered in PR #138 discussion

AlexKnauth:
When an `exn` is raised, use the check-info stack at the point the exception was raised, not the one at the point `(current-check-around)` was called.

This fixes the `standalone.rkt` failure about the presence of the `params` check-info if an error happens within the body of a check.

It doesn't really fix the `define-check infos are added before calling current-check-around` failure in `check-info-test.rkt` about the order of the `params` check-info relative to a custom check-info added by a user-defined `current-check-around`, instead it modifies the test to be less strict about the check-info/check-around order.

notjack:
> short answer: there needs to be a way for the code throwing the exception to communicate the check info stack *at the time the exception is created* to the `current-check-around` handler. If it’s a regular check failure created with `fail-check`, then `fail-check` will capture the current info stack and throw an instance of `exn:test:check` with the current continuation marks and the current info stack. But if it’s any *other* kind of exception, like a run-of-the-mill contract error, the info stack isn’t a field of the exception…. but the current info stack is a parameter, and parameters are stored in the continuation marks anyway, so it should be possible to get the `current-check-info` parameter’s value directly from the mark set.

### DIFF
--- a/rackunit-lib/rackunit/private/format.rkt
+++ b/rackunit-lib/rackunit/private/format.rkt
@@ -12,6 +12,16 @@
 (module+ for-test
   (provide display-check-info-stack))
 
+;; continuation-mark-set-parameter-value : Continuation-Mark-Set (Parameterof X) -> X
+(module continuation-mark-set-parameter-value racket/base
+  (require (only-in '#%paramz parameterization-key))
+  (provide continuation-mark-set-parameter-value)
+  (define (continuation-mark-set-parameter-value marks param)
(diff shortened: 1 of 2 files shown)


samth:
It looks like the CI problems still haven't been fixed, which is why the tests seem to be failing.

samth:
I've merged this. We'll see whether this breaks anything on DrDr, after which I'll merge this to the release branch.

PR #143, opened 2021-06-15
use paths of typed racket modules directly

capfredf:
related to https://github.com/racket/typed-racket/pull/1114/

### DIFF
--- a/rackunit-typed/rackunit/type-env-ext.rkt
+++ b/rackunit-typed/rackunit/type-env-ext.rkt
@@ -1,13 +1,14 @@
 #lang racket/base
 
-(require typed-racket/utils/utils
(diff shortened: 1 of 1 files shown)


sorawee:
I think the right fix for `check-info-test.rkt` and `standalone.rkt` is provided by @AlexKnauth in #142

capfredf:
Awesome. I will revert my fix.

sorawee:
Yes, I think it can be merged. It doesn't cause more problems than what we already have.

capfredf:
@sorawee can you help merge this pr?

sorawee:
Unfortunately no, I don't have a right to do that. Perhaps ask @samth or @jackfirth?

PR #148, opened 2021-08-04
reinstate info-value->string

aymanosman:
unbreaks rackunit-fancy-runner

See https://github.com/racket/rackunit/pull/140#issuecomment-883463970


aymanosman:
Seems to have fixed the bug @samth.

Screenshot 2021-08-04 at 23 19 05

samth:
Thanks!

PR #175, opened 2024-05-07
location: adjust path for editors

benknoble:
Running "raco test foo/bar/test.rkt" will cause Racket to change directory to "foo/bar" before invoking "test.rkt", and error messages are printed relative to (current-directory) like

    raco test: "foo/bar/test.rkt"
    --------------------
    fail
    FAILURE
    name:       check-equal?
    location:   test.rkt:5:0
    actual:     1
    expected:   2
    --------------------
    1/1 test failures

For an editor to consume this, it has to post-process the output to correlate "raco test: path" with "location: partial path" (which is difficult when running tests in parallel). This is not ideal; output with

    location:   foo/bar/test.rkt:5:0

is immediately useable by editors for jumping to the error.

Use the new current-test-invocation-directory to adjust the path better for editors, depending on compiler-lib 1.14. The output now looks like

    raco test: "foo/bar/test.rkt"
    --------------------
    fail
    FAILURE
    name:       check-equal?
    location:   foo/bar/test.rkt:5:0
    actual:     1
    expected:   2
    --------------------
    1/1 test failures

---

Depends on https://github.com/racket/racket/pull/4988.

### DIFF
--- a/rackunit-doc/rackunit/scribblings/utils.scrbl
+++ b/rackunit-doc/rackunit/scribblings/utils.scrbl
@@ -54,3 +54,8 @@ as a predicate and passed each export of the module. If @racket[skip] is
  Re-exports @raco-testing[test-log-enabled?] from @racketmodname[raco/testing].
  @history[#:added "1.1"
           #:changed "1.11" @elem{Allow any value for the parameter and coerce it to a boolean.}]}
+
+@defparam[current-test-invocation-directory test-invocation-directory (or/c #f path-string?) #:value #f]{
+Re-exports @raco-testing[current-test-invocation-directory] from @racketmodname[raco/testing].
+@history[#:added "1.2"]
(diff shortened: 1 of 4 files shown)


benknoble:
@jackfirth looks like the other PR was merged 🎉

samth:
It looks like this change broke the following test: http://drdr.racket-lang.org/66750/cs/racket/share/pkgs/rackunit-test/tests/rackunit/check-info-test.rkt

benknoble:
Thanks Sam; Matthew mentioned that to me, also for http://drdr.racket-lang.org/66749/cs/racket/share/pkgs/rackunit-test/tests/rackunit/check-info-test.rkt

I think what probably happened is the recent fixes to parallel modes caused the problem. I'll take a look as soon as I can.

PR #179, opened 2024-07-15
Show entire error message

rfindler:


rfindler:
I see that this is causing some test cases to fail, probably because they're checking the precise content of the error messages so probably the expected results need to be audited and adjusted.

PR #182, opened 2024-12-16
fix typing of check-regexp-match, make check return types Void

joergen7:
## Checklist

- [x] Bugfix
- [x] Feature
- [x] tests included
- [x] documentation

## Description of change

This pull request fixes several issues with `typed/rackunit`. Most importantly, it fixes the implementation and typing for `check-regexp-match` which is incorrectly typed as consuming three arguments but documented as consuming only two.

#### Observation

Consider the following check that uses an `input` argument adhering to `bytes?`:

    (check-regexp-match #rx"a+bba" #"aaaaaabba")

results in:

    --------------------
    FAILURE
     /home/jorgen/test.rkt:10:0
    name:       check-regexp-match
    location:   test.rkt:10:0
    params:     '(#rx"a+bba" #"aaaaaabba")
    --------------------

#### Expectation

Since `regexp-match` matches `check-regexp-match` should pass.

#### Execution

The implementation of `check-regexp-match` has been extended to not fail when valid input values other than `string?` are presented, these could now be `string?`, `bytes?`, `path?`, or `input-port?`.

#### Observation

Consider the following unit test in Typed Racket:

    (check-regexp-match #rx"a+bba" "aaaaaabba")

    Type Checker: could not apply function;
     wrong number of arguments provided
      expected: 3
      given: 2
      in: #%module-begin

does not type. Adding a bogus argument makes the check pass:

    (check-regexp-match #rx"a+bba" "aaaaaabba" #f)

Similarly, we can produce a failing test with a bogus argument:

    (check-regexp-match #rx"a+bba" "aaaaaabb" #f)

#### Expectation

Calling `check-regexp-match` should produce a passing test if only two arguments are provided.

#### Execution

This pull request removes the extra argument from the type specification of `check-regexp-match`.

#### Observation

The documentation of `check-regexp-match` provides the following example:

    (check-regexp-match "a+bba" "aaaaaabba")

In Typed Racket, however, this example fails not just because the type spec declares an extra argument (see above) but also because the first argument is restricted to be of type `Regexp` and here we provide a `String`.

#### Expectation

Since `check-regexp-match` uses `regexp-match` to create a unit test I would expect all arguments that can be passed to `regexp-match` to be valid also for `check-regexp-match`. Especially the above example from the documentation should type and pass.

#### Execution

This pull request extends the type spec for `check-regexp-match` to include `Regexp`, `Byte-Regexp`, `String`, and `Bytes` as valid regexp patterns. Also, it allows the input to be a `String`, `Bytes`, `Path`, or `Input-Port`.

#### Observation

Consider the following procedure:

    (: proc (-> Natural Void))
    (define (proc x)
      (check-eq? x 5))

This function fails to type raising:

    Type Checker: type mismatch
      expected: Void
      given: Any
      in: (define (proc x) (check-eq? x 5))

#### Expectation

The above function should type because `check-eq?`'s first argument is `Any` and `Natural` is a subtype of `Any`. The documenation says that `check-eq?` returns `void?`, so I expected the corresponding type to be `Void`. Setting the return type to `Any` works but is inadequate because the documentation says, it returns `void?`, not `any/c`.

#### Execution

All check's type specs have been updated to return `Void`.

### [doc] Update documentation

The documentation has been extended to account for the extended contract for `check-regexp-match`.

### DIFF
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -219,8 +219,8 @@ the check fails.
 
 }
 
-@defproc[(check-regexp-match (regexp (or/c regexp? string?))
-                             (string string?))
+@defproc[(check-regexp-match (regexp (or/c regexp? byte-regexp? string? bytes?))
+                             (string (or/c string? bytes? path? input-port?)))
          void?]{
 
 Checks that @racket[regexp] matches the @racket[string].
--- a/rackunit-doc/rackunit/scribblings/check.scrbl
+++ b/rackunit-doc/rackunit/scribblings/check.scrbl
@@ -219,7 +219,7 @@ the check fails.
 
 }
 
-@defproc[(check-regexp-match (regexp regexp?)
+@defproc[(check-regexp-match (regexp (or/c regexp? string?))
                              (string string?))
          void?]{
 
--- a/rackunit-lib/rackunit/private/util.rkt
+++ b/rackunit-lib/rackunit/private/util.rkt
@@ -84,5 +84,8 @@
       (test-case case-name case-body ...) ...))))
 
 (define-simple-check (check-regexp-match regex string)
-  (and (string? string) 
+  (and (or (string? string)
+           (bytes? string)
+           (path? string)
+           (input-port? string))
        (regexp-match regex string)))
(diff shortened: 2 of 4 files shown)


joergen7:
@samth a workflow needs approval of a maintainer.

sorawee:
I approved the workflow.

samth (approved):

PR #189, opened 2026-01-05
typed/private/utils and typed/private/rewriter are not used

ilya-klyuchnikov:
There imports are not used and can be safely removed.


samth:
Thanks!

